PR #9, opened 2021-11-16
Policy for accepting assembler optimisations

mattcaswell:
Details which branches and under what conditions assembler optimisations are acceptable.

### DIFF
--- a/policies/assembler.md
+++ b/policies/assembler.md
@@ -1,0 +1,19 @@
+Policy for Accepting Assembler Optimisations
+============================================
+
+New assembler optimisations for any algorithm having a C based implementation in
+any provider are always acceptable (subject to standard review procedures) for
+all platforms in the master branch.
+
+New assembler optimisations are never acceptable for any platform or provider in
+a stable release branch.
+
+Updates to existing assembler optimistations in a stable release branch are only
+acceptable where such updates would be allowed under the ["Stable Release
+Updates" policy][1].
+
+Where assembler optimisations are accetable they should be implemented using
+[perlasm][2].
+
+[1]: https://github.com/openssl/technical-policies/blob/master/policies/stable-release-updates.md
+[2]: https://github.com/openssl/openssl/blob/master/crypto/perlasm/README.md
--- a/policies/assembler.md
+++ b/policies/assembler.md
@@ -12,7 +12,7 @@ Updates to existing assembler optimistations in a stable release branch are only
 acceptable where such updates would be allowed under the ["Stable Release
 Updates" policy][1].
 
-Where assembler optimisations are accetable they should be implemented using
+Where assembler optimisations are acceptable they should be implemented using
 [perlasm][2].
 
 [1]: https://github.com/openssl/technical-policies/blob/master/policies/stable-release-updates.md


policies/assembler.md
```diff
@@ -0,0 +1,19 @@
...
+all platforms in the master branch.
+
+New assembler optimisations are never acceptable for any platform or provider in
+a stable release branch.
+
+Updates to existing assembler optimistations in a stable release branch are only
+acceptable where such updates would be allowed under the ["Stable Release
+Updates" policy][1].
+
+Where assembler optimisations are accetable they should be implemented using
```
t8m, line 15: Typo: acceptable

beldmit:
I'd add a proposal to prefer C intrinsic-based implementations to asm-based. But it probably should be a different document.

t8m:
> I'd add a proposal to prefer C intrinsic-based implementations to asm-based. But it probably should be a different document.

IMO not necessarily should. But why should be C intrinsic-based things be preferred?

beldmit:
We all know C reasonably well. At least I don't know asm at all. 

C intrinsics are just regular C functions, and the code using them can be normally reviewed.

t8m:
But C intrinsics are not supported by all C compilers. OpenSSL is not just gcc or clang C code. IMO that's the reason why perlasm was the preferred way.

beldmit:
They are supported at least by GCC, Clang, ICC, and MSVC. The proper #ifdef's may be a problem, though.

kroeckx:
> They are supported at least by GCC, Clang, ICC, and MSVC. The proper #ifdef's may be a problem, though.

We actually have some code making use of it, but that's:
#if !defined(OPENSSL_NO_ASM) && defined(__GNUC__)

It's assembler that's not doing anything, the assembler is
an empty string. And I doubt you can get that function to work
properly with MSVC without changes. There is a good reason for that
defined(__GNUC__).

paulidale (approved):
With the typo fix applied.

policies/assembler.md
```diff
@@ -0,0 +1,19 @@
...
+[perlasm][2].
+
+[1]: https://github.com/openssl/technical-policies/blob/master/policies/stable-release-updates.md
+[2]: https://github.com/openssl/openssl/blob/master/crypto/perlasm/README.md
```
levitte, line 19: That README only documents older methods that don't use for newer CPUs.  Perlasm also contains a number of translators that translate from GNU assembler to other assembler syntaxes.  The methods in this README are currently only used for x86

t-j-h: The readme needs updating - but that is separate from the policy which is pointing at the "right" place.

slontis (commented):
Apart from the typo this looks reasonable to me.

mattcaswell:
Typo fixed.

mattcaswell:
Vote for accepting the policy as of commit 50ec90f has now started.

mattcaswell:
Vote [+1]

paulidale:
Vote [+1]

t8m:
Vote [+1]

beldmit:
Vote [+1]

policies/assembler.md
```diff
@@ -0,0 +1,19 @@
+Policy for Accepting Assembler Optimisations
+============================================
+
+New assembler optimisations for any algorithm having a C based implementation in
+any provider are always acceptable (subject to standard review procedures) for
+all platforms in the master branch.
```
kroeckx, line 6: It's unclear what you mean with algorithm. The way things are implemented in assembler can be totally different than they are in C that I wouldn't call it the same algorithm. I don't see a need to restrict the assembler to do exactly the same as the C version. But maybe you just mean functionality and that all functionality should be available on all platforms?

It's also unclear what you mean with "all platforms in the master branch". We have various generic targets, and a new target can always be added in the same pull request.

And then there is the "in any provider". Does that include external providers?

policies/assembler.md
```diff
@@ -0,0 +1,19 @@
...
+
+New assembler optimisations are never acceptable for any platform or provider in
+a stable release branch.
```
kroeckx, line 9: Why does it have the "for any platform or provider" part?

policies/assembler.md
```diff
@@ -0,0 +1,19 @@
...
+
+Updates to existing assembler optimistations in a stable release branch are only
+acceptable where such updates would be allowed under the ["Stable Release
+Updates" policy][1].
+
+Where assembler optimisations are acceptable they should be implemented using
+[perlasm][2].
```
kroeckx, line 16: Why don't you just say: "Assembler optimisations should be implemented using [perlasm][2]."

policies/assembler.md
mattcaswell: To make it clear that it doesn't matter whether we're targeting a new platform or an existing platform, or whether we're targeting default or legacy providers - new assembler optimisations are never allowed. We could simplify the wording to just say "New assembler optimisations are never acceptable in a stable release branch" - but I don't think the current wording is wrong.

policies/assembler.md
mattcaswell: We could do. But I don't think the current wording is wrong.

mattcaswell:
IMO the current wording is clear enough. In principle I wouldn't object to modifying the wording to address some of @kroeckx comments - but I don't want to make amendments and restart the vote so I would prefer to continue the vote as is.

t-j-h:
Vote [+1]

levitte:
Vote:[+1]

slontis:
Vote: {+1]

t8m:
This got somehow neglected. I am closing the vote now.

t8m:
```
topic: Accept openssl/technical-policies PR#9 - Policy for accepting assembler optimisations
Proposed by Matt Caswell
Public: yes
opened: 2021-11-30
closed: 2022-01-03
accepted:  yes  (for: 8, against: 1, abstained: 0, not voted: 2)

  Dmitry     [+1]
  Matt       [+1]
  Pauli      [+1]
  Tim        [+1]
  Richard    [+1]
  Shane      [+1]
  Tomas      [+1]
  Kurt       [-1]
  Matthias   [  ]
  Nicola     [  ]
```

t8m:
The policy is now in place.

PR #10, opened 2021-11-28
Add Policy on API compatibility in minor releases

t8m:
This policy disallows all API changes in existing public API functions
and data in minor releases. Only API additions are allowed.

### DIFF
--- a/policies/api-compat.md
+++ b/policies/api-compat.md
@@ -1,0 +1,15 @@
+Policy on API compatibility in minor releases
+=============================================
+
+**No changes in existing public API functions and data are allowed.**
+
+Only API additions are allowed in minor releases.
+
+Although there might be some kinds of changes that are ABI compatible such
+as constification of parameters, changing a return value type from void
+to int to be able to indicate an error in the call, or changes of an API
+call from a macro to a function, these kinds of changes are allowed only in
+major releases.
+
+If necessary, a new API call can be added to implement the required changes in
+minor releases.
--- a/policies/api-compat.md
+++ b/policies/api-compat.md
@@ -1,6 +1,11 @@
 Policy on API compatibility in minor releases
 =============================================
 
+Public API of the OpenSSL libraries is defined as functions, macros, data
+structure declarations, typedefs, and data variables in header files in the
+`include/openssl` subdirectory of the source tree and `include/openssl`
+subdirectory of the build tree.
+
 **No changes to existing public API functions and data are permitted.** This
 includes, but is not limited to:
 
--- a/policies/api-compat.md
+++ b/policies/api-compat.md
@@ -1,15 +1,19 @@
 Policy on API compatibility in minor releases
 =============================================
 
-**No changes in existing public API functions and data are allowed.**
+**No changes to existing public API functions and data are permitted.** This
+includes, but is not limited to:
+
+- constification of arguments;
+- changing `void` returns to `int` returns;
+- changing a macro to a function and
+- corrections of spelling.
 
 Only API additions are allowed in minor releases.
 
-Although there might be some kinds of changes that are ABI compatible such
-as constification of parameters, changing a return value type from void
-to int to be able to indicate an error in the call, or changes of an API
-call from a macro to a function, these kinds of changes are allowed only in
-major releases.
+Although the changes as listed above might be regarded as ABI compatible, they
+cause various possible breakage when building applications depending on these
+APIs and thus they are not allowed in minor releases.
 
 If necessary, a new API call can be added to implement the required changes in
 minor releases.


policies/api-compat.md
```diff
@@ -0,0 +1,15 @@
...
+=============================================
+
+**No changes in existing public API functions and data are allowed.**
+
+Only API additions are allowed in minor releases.
+
+Although there might be some kinds of changes that are ABI compatible such
+as constification of parameters, changing a return value type from void
+to int to be able to indicate an error in the call, or changes of an API
+call from a macro to a function, these kinds of changes are allowed only in
```
paulidale, line 11: d/, these kinds of changes/

policies/api-compat.md
```diff
@@ -0,0 +1,15 @@
+Policy on API compatibility in minor releases
+=============================================
+
+**No changes in existing public API functions and data are allowed.**
+
+Only API additions are allowed in minor releases.
+
+Although there might be some kinds of changes that are ABI compatible such
```
paulidale, line 8: Does this mean constification is allowed?  Allowing it will break code that takes the address of a function.

if so, I'd suggest splitting this into two sentences.  The first being "Constification of parameters is allowed" or similar.

policies/api-compat.md
```diff
@@ -0,0 +1,15 @@
...
+as constification of parameters, changing a return value type from void
+to int to be able to indicate an error in the call, or changes of an API
```
paulidale, line 10: Drop the comma perhaps?

policies/api-compat.md
t8m: IMO this changes meaning.

policies/api-compat.md
t8m: No, nothing is allowed except additions of new things.

paulidale: It reads as if some changes are permitted.

How about:
> No changes to any existing API are permitted.  This includes, but is not limited to:
> * constification of arguments;
> * changing void returns to int returns;
> * changing a macro to a function and
> * corrections of spelling.

t8m: OK, done.

policies/api-compat.md
```diff
@@ -0,0 +1,19 @@
+Policy on API compatibility in minor releases
+=============================================
+
+**No changes to existing public API functions and data are permitted.** This
```
mattcaswell, line 4: Do we need a definition of what constitutes a "public API function" or "data"?

t8m: I don't know. It should obviously be anything declared in the header files in `include/openssl`. Anything else?

mattcaswell: Probably that is sufficient. I do think it needs to be explicitly said though because "public API functions" could reasonably be interpreted as anything in the docs (which is not what is intended I think).

paulidale: ~We defined public API in one of the other documents I think.~  Correction: we don't define it.

It might be worthwhile thinking about having a policy definitions document which other policies can reference.  We definitely do not want to end up defining a term in more than one place.

mattcaswell: > We definitely do not want to end up defining a term in more than one place.

True. Especially if they're not the same definition! :-)

t8m: Sooo.... what to do with this policy - should we for now add the definition of public API here and later we do a separate policy document with definitions?

mattcaswell: I'd be happy with adding it here for now and fixing it later. The stable release updates policy also adds a lot of definitions - so getting a standard "definitions" policy will at least need to make updates to that.

t8m:
Added definition of public API. Suggestions on how to improve it welcome.

policies/api-compat.md
kroeckx: At least one of the definitions is what is a minor release. We might want to have a policy for patch level releases, assuming it's more restrictive than that of a minor release.

policies/api-compat.md
```diff
@@ -0,0 +1,24 @@
...
+
+Public API of the OpenSSL libraries is defined as functions, macros, data
+structure declarations, typedefs, and data variables in header files in the
+`include/openssl` subdirectory of the source tree and `include/openssl`
+subdirectory of the build tree.
+
+**No changes to existing public API functions and data are permitted.** This
+includes, but is not limited to:
+
+- constification of arguments;
```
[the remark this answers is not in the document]
t8m: Once the function is used as a function pointer, even constification of arguments changes the type of the function pointer.

policies/api-compat.md
t8m: Actually there is no difference in regards to this proposal on what can be done within a minor release and in between minor releases. Any non-additions can be done only for a major release.

levitte:
Vote: [+1]

mattcaswell:
Vote [+1}

t-j-h:
Vote: [+1]

t8m:
Vote for accepting the policy at the commit id 2092743 has started.

t8m:
Vote: [+1]

kroeckx:
Vote: [+1]

paulidale:
Vote: [+1]

beldmit:
Vote: [+1]

slontis:
Vote: [+1]

t8m:
```
topic: Accept openssl/technical-policies PR#10 - Add Policy on API compatibility in minor releases
Proposed by Tomáš Mráz
Public: yes
opened: 2021-12-14
closed: 2022-01-03
accepted:  yes  (for: 8, against: 0, abstained: 0, not voted: 1)

  Dmitry     [+1]
  Matt       [+1]
  Pauli      [+1]
  Tim        [+1]
  Richard    [+1]
  Shane      [+1]
  Tomas      [+1]
  Kurt       [+1]
  Matthias   [  ]
  Nicola     [  ]
```

t8m:
The policy is now in place.

PR #13, opened 2021-12-13
Add a Testing Policy

mattcaswell:


policies/testing.md
```diff
@@ -0,0 +1,42 @@
+Testing Policy
+==============
+
+This applies to all stable and development branches of the main code repository.
+Except where noted below:
+
+- All Pull Requests adding new functionality to the applications, libraries,
+providers or engines must include suitable tests.
+- All Pull Requests fixing a functional behavior defect in the applications,
+libraries, providers or engines must include a test for that defect.
```
kroeckx, line 10: Do have or need a definition functional behaviour? Is any bugfix a functional behaviour change?

kroeckx:
This policy should probably include adding fuzzers.

policies/testing.md
```diff
@@ -0,0 +1,42 @@
...
+libraries, providers or engines do not require tests to be added. For example
+the following types of changes do not require tests:
+ - Changes to comments, formatting, internal naming or similar
+ - Fixes for performance
+ - Refactoring
+
+To be explicit the above statement means that tests are not required for changes
+in:
+ - Documentation (including CHANGES.md/NEWS.md)
+ - The test suite
```
levitte, line 22: Hmmm...  we have a test utilities library, and we *do* have a test for it: `test/recipes/01-test_test.t` and `test/test_test.c`...

paulidale: These tests haven't aged well.  There have been a lot of things added to the test framework which aren't covered.

mattcaswell:
- Include wording about difficult to reproduce error conditions
- Tests could be added in different PRs
- Fuzz testing for substantial features

policies/testing.md
```diff
@@ -0,0 +1,42 @@
...
+- All Pull Requests adding new functionality to the applications, libraries,
+providers or engines must include suitable tests.
+- All Pull Requests fixing a functional behavior defect in the applications,
+libraries, providers or engines must include a test for that defect.
+
+Pull Requests that do not change the functional behaviour of the applications,
+libraries, providers or engines do not require tests to be added. For example
+the following types of changes do not require tests:
+ - Changes to comments, formatting, internal naming or similar
+ - Fixes for performance
```
kroeckx, line 16: If it's about performance, and it was not something we monitored under the performance testing before, should we not add it to that? Clearly someone thought it was important enough to fix it so that we want to monitor it for regressions.

policies/testing.md
```diff
@@ -0,0 +1,42 @@
...
+ - Refactoring
```
kroeckx, line 17: Refactoring seems to be a good time to check that corner cases are tested. It might not need new tests, but it seems to be a good time to check that there are enough tests.

policies/testing.md
```diff
@@ -0,0 +1,42 @@
...
+
+Performance Testing
+-------------------
+
+Performance testing should be performed automatically via CI on a regular basis
+for certain defined components.
+
+Components that should be performance tested via CI include:
+ - openssl speed for all supported algorithms
+ - openssl s_time
```
kroeckx, line 38: This seems to mention implementation details rather than what we want to do. I prefer that the policy juts mentions what kind of things should be tested.

policies/testing.md
```diff
@@ -0,0 +1,42 @@
...
+ - Build system
+ - Demos
+
+Pull Requests that only affect minor cosmetic output do not require tests
+
+Performance Testing
+-------------------
+
+Performance testing should be performed automatically via CI on a regular basis
+for certain defined components.
```
kroeckx, line 34: It says on regular basis. Does that mean that it's not run for each pull request?

policies/testing.md
mattcaswell: I've added a paragraph attempting to explain what is meant by this.

policies/testing.md
mattcaswell: The test policy says that tests are not required....but it doesn't prevent us from adding them if we want them anyway.

policies/testing.md
mattcaswell: I've added a non-mandatory "Recommendations" section and added recommendation to consider whether a performance test is appropriate.

policies/testing.md
mattcaswell: I've added a recommendation to this effect.

policies/testing.md
mattcaswell: I've reworded this and dropped specific mention of speed and s_time

policies/testing.md
mattcaswell: I wasn't imagining this being run on every pull request.

mattcaswell:
> Include wording about difficult to reproduce error conditions
Tests could be added in different PRs
Fuzz testing for substantial features

Fixup pushed addressing the above items as well as the feedback so far.

policies/testing.md
```diff
@@ -0,0 +1,71 @@
...
+Within this policy "functional behaviour" means what the system does, i.e.
+given a set of inputs it will produce a set of outputs. This does not include
+how the system does it. For example refactoring or performance improvements do
+not affect functional behaviour.
+
+Except where noted below:
+
+- All Pull Requests adding new functionality to the applications, libraries,
+providers or engines must include suitable tests.
+- All Pull Requests fixing a functional behavior defect in the applications,
```
t8m, line 15: I think it is overly strict to require a test for every small bug fix. Especially for the cases where writing the test would require adding significant infrastructure (tests depending on triggering a malloc failure, tests having to add a new provider implementation or other cases). IMO the policy should be at least somewhat more lenient such as allowing the reviewer(s) to give an exception for this rule.

t8m: Ah, I see you've added this already below. OK then.

t8m (changes requested):

policies/testing.md
mattcaswell: I've only added an exception where the reviewers can agree for "difficult to reproduce error conditions". Should there be a broader exception?

t8m: Maybe add something like that reviewers can give an exception where writing the test would require significantly higher effort than writing the bug fix.

mattcaswell: I'm not sure "signficiantly higher" is quite the right bar. Its not that uncommon that writing the test is significantly more work than fixing the bug. I don't think that should necessarily be an excuse to not write a test. However, I wouldn't expect someone to have to go and write a whole new testing framework to test specific classes of things. I've made an update that says "disproportionately more effort".

t8m (approved):

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+ - Include files
+ - Build system
+ - Demos
+
+Pull Requests that only affect minor cosmetic output do not require tests.
+
+By agreement of the Pull Request reviewers it is acceptable for the tests to be
+added via a different Pull Request to the main Pull Request.
+
+By agreement of the Pull Request reviewers a test may be omitted where writing
```
t-j-h, line 39: Delete "By agreement of the Pull Request reviewers "

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+ - Documentation (including CHANGES.md/NEWS.md)
+ - The test suite
+ - perl utilities
+ - Include files
+ - Build system
+ - Demos
+
+Pull Requests that only affect minor cosmetic output do not require tests.
+
+By agreement of the Pull Request reviewers it is acceptable for the tests to be
```
t-j-h, line 36: Delete "By agreement of the Pull Request reviewers "

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+Performance testing should be performed automatically via CI on a regular basis
+for certain components as defined by the OTC from time to time.
+
+Examples of performance testing that should be considered include:
+- Individual algorithm performance operating over different input sizes
+- SSL/TLS handshake time over multiple handshakes and for different protocol
+versions and resumption/non-resumption handshakes
+
+Performance figures should be collected and tracked over time. Alerts should be
```
t-j-h, line 54: Remove "Alerts should be sent" - that's operational and not policy.

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+sent if performance drops below an agreed threshold versus an agreed baseline
+as set from time to time by the OTC.
```
t-j-h, line 56: Does not make sense to say "from time to time" ... can delete that

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+---------------
+
+This section makes recommendations that are not mandatory but should be
+considered.
+
+Pull requests that implement significant new functionality should consider
+whether fuzz tests should be added.
+
+As per the policy wording above, pull requests that implement refactoring are
+not required to add new tests. However refactoring is a good time to check that
```
t-j-h, line 68: It is "before refactoring" that you want to make sure the tests are decent - not during, not after.

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+Within this policy "functional behaviour" means what the system does, i.e.
+given a set of inputs it will produce a set of outputs. This does not include
+how the system does it. For example refactoring or performance improvements do
+not affect functional behaviour.
+
+Except where noted below:
+
+- All Pull Requests adding new functionality to the applications, libraries,
+providers or engines must include suitable tests.
+- All Pull Requests fixing a functional behavior defect in the applications,
```
t-j-h, line 15: behavior -> behaviour - be consistent throughout the document

policies/testing.md
kroeckx: So any change that you can observe as caller is a function behaviour change.

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+ - The test suite
+ - perl utilities
+ - Include files
+ - Build system
+ - Demos
+
+Pull Requests that only affect minor cosmetic output do not require tests.
+
+By agreement of the Pull Request reviewers it is acceptable for the tests to be
+added via a different Pull Request to the main Pull Request.
```
kroeckx, line 37: Do we want to enforce that it gets written? And so that the pull requests that makes the behavioural changes can only get merged as the same time as the one testing it?

policies/testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+By agreement of the Pull Request reviewers a test may be omitted where writing
+the test would result in disproportionately more effort than writing the code
+being tested. For example, difficult to reproduce error conditions.
+
+Performance Testing
+-------------------
+
+Performance testing should be performed automatically via CI on a regular basis
+for certain components as defined by the OTC from time to time.
```
kroeckx, line 47: So that's not part of a pull request?

t8m: Performance testing will have to be implemented via special CI runs being run on our own HW. I am not sure we want to run these on each and every PR.

policies/testing.md
mattcaswell: Added "However before refactoring..."

policies/testing.md
mattcaswell: I'm not sure. The current text doesn't require it. I'd like to hear the OTC view on this.

policies/testing.md
mattcaswell: Correct - not part of the pull request (I think the performance tests will be too "heavy" to run on every PR).

mattcaswell:
Fixup applied addressing the latest feedback.

policies/testing.md
t8m: **OTC does not see the need to change the policy in this way for now.**

levitte:
Formatting suggestion:

``` diff
diff --git a/policies/testing.md b/policies/testing.md
index 13b2096..fcc8ae1 100644
--- a/policies/testing.md
+++ b/policies/testing.md
@@ -11,25 +11,27 @@ not affect functional behaviour.
 Except where noted below:
 
 - All Pull Requests adding new functionality to the applications, libraries,
-providers or engines must include suitable tests.
+  providers or engines must include suitable tests.
 - All Pull Requests fixing a functional behaviour defect in the applications,
-libraries, providers or engines must include a test for that defect.
+  libraries, providers or engines must include a test for that defect.
 
 Pull Requests that do not change the functional behaviour of the applications,
 libraries, providers or engines do not require tests to be added. For example
 the following types of changes do not require tests:
- - Changes to comments, formatting, internal naming or similar
- - Fixes for performance
- - Refactoring
+
+- Changes to comments, formatting, internal naming or similar
+- Fixes for performance
+- Refactoring
 
 To be explicit the above statement means that tests are not required for changes
 in:
- - Documentation (including CHANGES.md/NEWS.md)
- - The test suite
- - perl utilities
- - Include files
- - Build system
- - Demos
+
+- Documentation (including CHANGES.md/NEWS.md)
+- The test suite
+- perl utilities
+- Include files
+- Build system
+- Demos
 
 Pull Requests that only affect minor cosmetic output do not require tests.
 
@@ -47,9 +49,10 @@ Performance testing should be performed automatically via CI on a regular basis
 for certain components as defined by the OTC.
 
 Examples of performance testing that should be considered include:
+
 - Individual algorithm performance operating over different input sizes
 - SSL/TLS handshake time over multiple handshakes and for different protocol
-versions and resumption/non-resumption handshakes
+  versions and resumption/non-resumption handshakes
 
 Performance figures should be collected and tracked over time.
 
```

mattcaswell:
> Formatting suggestion:

Applied.

mattcaswell:
Vote for accepting the policy at the commit id ac9026dcba7d has started.

mattcaswell:
Vote: [+1]

t-j-h:
Vote: [+1]

paulidale:
Vote: [+1]

mattcaswell:
> While I cannot vote, I would like to ask a question: For functionality that is specific to a platform and cannot be generally tested in CI/CD systems, how will this be handled in relation to this policy?

This policy is mostly about what things should and should not have tests and has very little to do with CI. There is no distinction made about platform specific functionality. All functionality (except where the various caveats apply) should have tests (as run by "make test"). We would expect those tests to pass on all platforms. If there is no CI available for a given platform, then we won't know about test failures until someone tells us about them. In the event that there is functionality that is only available on a subset of platforms then I would expect the tests to be skipped on platforms where they are not relevant.

The platform policy is much more relevant when talking about the availability of CI (https://www.openssl.org/policies/platformpolicy.html). This tells us what platforms will have CI and which ones will not.

slontis:
Vote: [+1]

t8m:
Vote: [+1]

beldmit:
Vote: [+1]

kroeckx:
I'm voting -1 on this. I agree in principle on the direction, but I have a problem with when tests need to be added. It's my understanding that any bug fix is a functional behaviour change, and so will require a test. That sounds nice, but not all bug fixes need testing, and this policy will probably be violated on very regular basis.

mspncp:
Vote: [+0]

I share @kroeckx's concerns that requiring tests for all bugfixes might be a little bit too strict. Do we really want a missing testcase to prevent a bugfix to get merged, even if it obviously fixes the defect?

Maybe it would be better to weaken the **must** to a **shall** in the bugfix case?

- All Pull Requests adding new functionality to the applications, libraries,
  providers or engines **must** include suitable tests.
- All Pull Requests fixing a functional behaviour defect in the applications,
  libraries, providers or engines **shall** include a test for that defect.

romen:
Vote: [+1]

mattcaswell:
Closing this vote now:

````
Topic: Accept openssl/technical-policies PR#13 - Testing policy as of
commit ac9026d. This will become an official OTC policy.
Proposed by: Matt Caswell
Issue link: https://github.com/openssl/technical-policies/pull/13
Public: yes
Opened: 2022-01-25
Closed: 2022-02-01
Accepted:  yes  (for: 7, against: 1, abstained: 1, not voted: 1)

  Dmitry     [+1]
  Matt       [+1]
  Pauli      [+1]
  Tim        [+1]
  Richard    [  ]
  Shane      [+1]
  Tomas      [+1]
  Kurt       [-1]
  Matthias   [+0]
  Nicola     [+1]
````

levitte:
vote: [+0]

PR #17, opened 2022-01-12
Make the voting policy match reality

mattcaswell:
The filenames for votes specified by the policy, and the filenames actually
used are slightly different. Bring the policy into line with reality.

### DIFF
--- a/policies/voting-procedure.md
+++ b/policies/voting-procedure.md
@@ -56,7 +56,7 @@ a vote against, `[+0]` an abstention with an inclination in favour,
 `[ 0]` a neutral abstention, `[-0]` an abstention with an inclination
 against, and `[  ]` meaning not voted.
 
-The vote files are named `yyyymmdd-vote-short-id.txt` where the `yyyymmdd`
+The vote files are named `vote-yyyymmdd-vote-short-id.txt` where the `yyyymmdd`
 is the date when the vote was proposed and the `vote-short-id` is a short
 mnemonic identifier of the vote such as `voting-procedure` or `accept-pr-1234`
 or similar.


t8m:
Should we just rename the files?

mattcaswell:
> Should we just rename the files?

I thought about that, but I think in a previous meeting someone said that the current naming scheme was actually more informative.

t8m:
OK then.

Now the question is - do we have to wait two weeks to start vote for this trivial policy change? :grin:

mattcaswell:
> Now the question is - do we have to wait two weeks to start vote for this trivial policy change?

This is not an urgent change, so it can wait if needs be. However we might actually want to consider whether it is appropriate to have a "fast track" path for relatively trivial policy changes.

t-j-h:
There should **never** be a context where we need to fast track a policy change - we can always vote exceptions to a policy on a case by case basis - but updates of policies should **always** go through the visible PRs and a time period allowing for anyone that wants to comment to comment on them.

levitte:
> There should **never** be a context where we need to fast track a policy change - we can always vote exceptions to a policy on a case by case basis - but updates of policies should **always** go through the visible PRs and a time period allowing for anyone that wants to comment to comment on them.

Maybe we should make that an actual policy.  I've been looking around to remind myself if we already have that sort of policy, but could find none of the sort.  I may have stared myself blind, so please remind me where to look if there is something like this somewhere.

t8m:
@levitte it is explicitly mentioned here: https://github.com/openssl/technical-policies/blame/master/policies/policy-change-process.md#L34

levitte:
> @levitte it is explicitly mentioned here: https://github.com/openssl/technical-policies/blame/master/policies/policy-change-process.md#L34

Ah, yeah...  but that doesn't say when voting may start or not, at least the way I read it.

t8m:
Yeah, the wording is less then ideal. It was meant to be implied that the two weeks are for the Review Process phase of the policy change process.

t8m:
There is also https://github.com/openssl/technical-policies/blame/master/policies/policy-change-process.md#L55

levitte:
> There is also https://github.com/openssl/technical-policies/blame/master/policies/policy-change-process.md#L55

Thanks

mattcaswell:
Vote for accepting the policy at the commit id 46fbe55 has started.

mattcaswell:
Vote: [+1]

t8m:
Vote: [+1]

paulidale:
Vote: [+1]

beldmit:
Vote: [+1]

kroeckx:
Vote: 0

mspncp:
Vote: [+1]

slontis:
Vote: [+1]

romen:
Vote: [+1]

t-j-h:
Vote: [+1]

levitte:
Vote: +1

mattcaswell:
Closing the vote now:

````
Topic: Accept openssl/technical-policies PR#17 - Voting policy update as of
commit 46fbe55. This will become an official OTC policy.
Proposed by: Matt Caswell
Issue link: https://github.com/openssl/technical-policies/pull/17
Public: yes
Opened: 2022-01-28
Closed: 2022-02-01
Accepted:  yes  (for: 9, against: 0, abstained: 1, not voted: 0)

  Dmitry     [+1]
  Matt       [+1]
  Pauli      [+1]
  Tim        [+1]
  Richard    [+1]
  Shane      [+1]
  Tomas      [+1]
  Kurt       [ 0]
  Matthias   [+1]
  Nicola     [+1]
````

PR #18, opened 2022-01-18
Documentation policy

paulidale:
First draft

### DIFF
--- a/policies/documentation-policy.md
+++ b/policies/documentation-policy.md
@@ -1,0 +1,122 @@
+# OpenSSL documentation policy
+
+This document describes the code documentation and commenting
+requirements for the OpenSSL project.
+
+The project's documentation is all about making the libraries and tools
+more accessible to our users and making the code more maintainable.
+This policy applies to new submissions, existing code is below par and will
+be gradually improved.
+
+## Chapter 1: Command line commands and arguments
+
+All new or modified arguments to the commands must be documented in the
+`doc/man1` directory.  This documentation is in _POD_ format.
+
+All new commands must be documented in the `doc/man1` directory.  This
+documentation is in _POD_ format.
+
+## Chapter 2: Public symbols in the libraries
+
+All new public symbols must be documented in a _POD_ manual page in the
+`doc/man3` directory.  This includes both macros and functions.
+
+The allowed exceptions are:
+
+- guard macros preventing a header file being included twice
+- new OPENSSL_NO_DEPRECATED_ macros
+- new symbols generated automatically via `make update` (errors, objects, etc)
+
+## Chapter 3: Overviews, conventions, et al
+
+Where additional user facing information is required, it should be
+included in the `doc/man7` section.  This includes, but is not limited to:
+
+- descriptions of data structures and their usage
+- provider calls and call backs
+- algorithm descriptions and parameters
+- descriptions of architectural decisions
+
+## <a name="fsgm"></a>Chapter 4: Internal functions, structures, globals and macros
+
+Internal functions and macros are those defined in the `include/crypto`
+and `include/internal` directories.
+
+These should all be documented at the point of implementation.  A comment
+describing their purpose and the input and output arguments and return value
+for functions suffices.
+
+For _trivial_ items, where their operation is obvious from their
+implementation, the documentation requirement is not mandated.  The following
+are generally representative of trivial items, however it is quite
+possible for any of these to be non-trivial in specific instances and
+therefore require documentation:
+
+- OSSL_DISPATCH tables
+- upref functions
+- free functions
+- simple getter/setter functions
+- wrappers for other functions (a function that calls a more recent `_ex`
+  variant or a group of functions that call a common internal routine)
+
+For structures, each of the fields should be commented stating its
+purpose.  Again, a _trivial_ exception applies where the purpose is
+obvious.  Some representative examples:
+
+- `OSSL_LIB_CTX *ctx;` where there is only one library context referenced in
+  the structure.
+- `struct *next;` in a linked list implementation.
+- `CRYPTO_REF_COUNT refcnt;`
+
+## Chapter 5: Static functions and globals and local structures and macros
+
+These are functions, structures, globals and macros local to a specific
+C file or defined for a single directory as part of a local .h file.
+
+These should all be documented at the point of implementation.  Follow the
+rules and exceptions listed [above](#fsgm).  In some cases slightly more
+leniency with respect to _trivial_ can be tolerated.
+
+## Chapter 6: Code comments
+
+Comments in code should explain what the algorithm is doing and why.
+It should not parrot the effect of each statement.  As the complexity of
+the code increases, the size and detail of comments should also increase.
+
+## Chapter 7: Assembly code
+
+Assembly code should include a good description of the algorithm and
+approach being used.  This should be followed by a performance comparison
+and then the assembly code itself.  The assembly code should be well
+commented, but it isn't necessary to comment every line.  A comment
+describing each block of code suffices.
+
+There are no _trivial_ exceptions for assembly code.
+
+## Chapter 8: Configure options
+
+New options added to the configuration scripts must be documented in the
+`INSTALL.md` file.
+
+## Chapter 9: Changes and news
+
+Significant modifications should be documented in the `CHANGES.md` file.
+
+Very significant features and changes should be documented in the `NEWS.md`
+file.
+
+In both cases, the added note should be short and to the point.
+
+## Chapter 10: Automated sanity checking
+
+The `make doc-nits` command should be run before submitting a pull
+request and any problems it locates must be addressed.
+
+## Chapter 11: Common sense
+
+Comments in the code are to improve readability and comprehension.
+Where the code is obvious, there is no need to include a comment.
+However, common sense applies.  Always err in favour of including more
+comments than less or none.  Code that you've just written that is
+_obvious_ will not necessarily be to someone else two years later.
+
--- a/votes/vote-20220208-documentation-policy.txt
+++ b/votes/vote-20220208-documentation-policy.txt
@@ -1,0 +1,19 @@
+Topic: Accept openssl/technical-policies PR#18 - Documentation policy as of
+commit 49f3b24. This will become an official OTC policy.
+Proposed by: Pauli
+Issue link: https://github.com/openssl/technical-policies/pull/18
+Public: yes
+Opened: 2022-02-03
+Closed: 2022-02-08
+Accepted:  yes  (for: 7, against: 1, abstained: 0, not voted: 2)
+
+  Dmitry     [+1]
+  Matt       [+1]
+  Pauli      [+1]
+  Tim        [+1]
+  Richard    [  ]
+  Shane      [+1]
+  Tomas      [+1]
+  Kurt       [-1]
+  Matthias   [+1]
+  Nicola     [  ]


policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+definition, the documentation requirement is not required.  The following
+are generally representative of trivial items, however it is quite
+possible for any of these to be non-trivial in specific instances and
+therefore require documentation:
+
+- OSSL_DISPATCH tables
+- method structures
+- upref functions
+- free functions
+- simple getter/setter functions
```
t-j-h, line 47: Where structures are opaque the "fields" of the underlying structure need to be documented and in the setters/getters is where that documentation will occur. Unless you plan to require documenting this somewhere else. So I would **not** include setter/getter in the list of normally not documented.

paulidale: Hmmmm, there are a number of _get0_name, _get_number, get_size where a comment isn't adding anything meaningful.  There are others where there should be some documentation.

t8m: As I understand it this is in regards to internal documentation so the fields are not really opaque here and so it should be sufficient to document the structure members itself via comments. For public getters/setters the documentation is required.

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
+# OpenSSL documentation policy
+
+This document describes the code documentation and commenting
+requirements for the OpenSSL project.
```
t8m, line 4: As we certainly do not fully adhere to this policy with the current code base we should state that the policy in full applies only to new additions to the code. I.E. we do not require authors of trivial bugfix PRs to add a documentation of the internal function being fixed if it is missing and we do not plan/we are not required to make the existing code base to comply with the policy ASAP. The compliance will be just incrementally improved.

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+
+The project's documentation is all about making the libraries and tools
+more accessible to our users.
```
mattcaswell, line 7: Since this policy also covers internal documentation/code comments etc, it is additionally about code maintenance

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+- guard macros preventing a header file being included twice
+- new OPENSSL_NO_DEPRECATED_ macros
+- new symbols generated automatically via `make update` (errors, objects, etc)
+
+## Chapter 3: Internal functions, structures, globals and macros
+
+Internal functions and macros are those defined in the `include/crypto`
+and `include/internal` directories.
+
+These should all be documented at the point of definition.  A comment
```
mattcaswell, line 33: I'm not sure what "definition" actually means here. Can we make a distinction between declaration (header file) and implementation (C source file)?

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+
+For structures, each of the fields should be commented stating its
+purpose.  Again, a _trivial_ exception applies where the purpose is
+obvious.  Some representative examples:
+
+- `OSSL_LIB_CTX *ctx;` where there is only one library context referenced in
+  the structure.
+- `struct *next;` in a linked list implementation.
+- `CRYPTO_REF_COUNT refcnt;`
+- `CRYPTO_RWLOCK *lock;` where there is only one lock in the structure.
```
mattcaswell, line 59: I'm not sure the lock is trivial. Some kind of definition of the scope of things that should be locked should be provided. E.g. some locks only cover specific fields or use cases.

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+All new or modified arguments to the commands must be documented in the
+`doc/man1` directory.  This documentation is in _POD_ format.
+
+All new commands must be documented in the `doc/man1` directory.  This
+documentation is in _POD_ format.
+
+## Chapter 2: Public symbols in the libraries
+
+All new public symbols must be documented in a _POD_ manual page in the
+`doc/man3` directory.  This includes both macros and functions.
```
mattcaswell, line 20: Should we say something about man7 documentation?

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+commented, but it isn't necessary to comment every line.  A comment
+describing each block of code suffices.
+
+There are no _trivial_ exceptions for assembly code.
+
+## Chapter 7: Automated sanity checking
+
+The `make doc-nits` command should be run before submitting a pull
+request and any problems it locates must be addressed.
+
```
mattcaswell, line 90: This policy doesn't seem to cover:

- New Configure options (i.e. INSTALL.md)
- CHANGES.md and NEWS.md

mattcaswell (changes requested):

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+- new symbols generated automatically via `make update` (errors, objects, etc)
+
+## Chapter 3: Internal functions, structures, globals and macros
+
+Internal functions and macros are those defined in the `include/crypto`
+and `include/internal` directories.
+
+These should all be documented at the point of definition.  A comment
+describing their purpose and the input and output arguments for functions
+suffices.
```
kroeckx, line 35: and return values

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+
+For _trivial_ items, where their operation is obvious from their
+definition, the documentation requirement is not required.  The following
+are generally representative of trivial items, however it is quite
+possible for any of these to be non-trivial in specific instances and
+therefore require documentation:
+
+- OSSL_DISPATCH tables
+- method structures
```
kroeckx, line 44: What about instance of such a method structure? It's now generally just a list of function names that's filled in. Should we document (using comment) which fields are being set?

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+- upref functions
+- free functions
```
kroeckx, line 46: I assume alloc could get mentioned too.

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+- simple getter/setter functions
+- wrappers for other functions (a function that calls a more recent `_ex`
+  variant)
```
kroeckx, line 49: Do we do this for internal functions?

policies/documentation-policy.md
```diff
@@ -0,0 +1,91 @@
...
+These are functions, structures, globals and macros local to a specific
+C file or defined for a single directory as part of a local .h file.
+
+These should all be documented at the point of definition.  Follow the
+rules and exceptions listed in Chapter 3.  In some cases slightly more
+leniency with respect to _trivial_ can be tolerated.
+
+## Chapter 5: Code comments
+
+Comments in code should explain what the algorithm is doing, not each
```
kroeckx, line 72: I think why is more important that what. What it's doing is for when it's not obvious what it's doing.

policies/documentation-policy.md
paulidale: I was trying to.  Implementation is a better description.

policies/documentation-policy.md
paulidale: Which wouldn't meet the trivial criteria :)

policies/documentation-policy.md
paulidale: Allocation functions are generally more complex than free.  I think some documentation about them is worthwhile.  But yes, there is no need for XXX_new() to have a leading comment `/* allocates a new XXX */`.

policies/documentation-policy.md
paulidale: I've added a clarification.

policies/documentation-policy.md
paulidale: I meant to but forgot about it.  I'll add sections for these.

policies/documentation-policy.md
paulidale: I think the same would apply.  I've clarified.

These are examples where commenting isn't required -- I'd like to leave the list short and understated rather than trying to cover everything.

policies/documentation-policy.md
paulidale: Good point.

paulidale:
Feedback addressed (hopefully).

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+
+Internal functions and macros are those defined in the `include/crypto`
+and `include/internal` directories.
+
+These should all be documented at the point of implementation.  A comment
+describing their purpose and the input and output arguments and return value
+for functions suffices.
+
+For _trivial_ items, where their operation is obvious from their
+implementation, the documentation requirement is not required.  The following
```
mattcaswell, line 50: "the requirement is not required" sounds a bit odd

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+
+For structures, each of the fields should be commented stating its
+purpose.  Again, a _trivial_ exception applies where the purpose is
+obvious.  Some representative examples:
+
+- `OSSL_LIB_CTX *ctx;` where there is only one library context referenced in
+  the structure.
+- `struct *next;` in a linked list implementation.
+- `CRYPTO_REF_COUNT refcnt;`
+- `CRYPTO_RWLOCK *lock;` where there is only one lock in the structure and
```
mattcaswell, line 70: I actually think a lock is virtually never trivial. Even in the case where there is only one and it locks the whole structure, I think it is necessary to explicitly state that it locks the whole structure.

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+  it locks the entire structure.
+
+## Chapter 5: Static functions and globals and local structures and macros
+
+These are functions, structures, globals and macros local to a specific
+C file or defined for a single directory as part of a local .h file.
+
+These should all be documented at the point of implementation.  Follow the
+rules and exceptions listed in Chapter 3.  In some cases slightly more
```
mattcaswell, line 79: Do you mean chapter 4 here?

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+approach being used.  This should be followed by a performance comparison
+and then the assembly code itself.  The assembly code should be well
+commented, but it isn't necessary to comment every line.  A comment
+describing each block of code suffices.
+
+There are no _trivial_ exceptions for assembly code.
+
+## Chapter 8: Configure options
+
+New options added to the configuration scripts must be documents in the
```
mattcaswell, line 100: s/documents/documented/

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+`INSTALL.md` file.
+
+## Chapter 9: Changes and news
+
+Non-trivial modification should be documented in the `CHANGES.md` file.
```
mattcaswell, line 105: "non-trivial" is quite a different bar to what we do at the moment. We generally reserve CHANGES.md updates to significant changes.

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+
+Significant features and changes should be documented in the `NEWS.md` file.
```
mattcaswell, line 107: We generally reserve this for very significant changes.

policies/documentation-policy.md
```diff
@@ -0,0 +1,123 @@
...
+The `make doc-nits` command should be run before submitting a pull
+request and any problems it locates must be addressed.
+
+## Chapter 11: Common sense
+
+Comments in the code are to improve readability and comprehension.
+Where the code is obvious, there is no need to include a comment.
+However, common sense applies.  Always err in favour of including more
+comments than less or none.  Code that you've just written that is
+_obvious_ will not necessarily be to someone else two years later.
```
mattcaswell, line 122: This seems to be entirely about code comments. Should it actually be incorporated into chapter 6?

mattcaswell (changes requested):

policies/documentation-policy.md
paulidale: Removed.

policies/documentation-policy.md
paulidale: Yes, things were renumbered and I missed this.  I'll switch to a link to the right spot and omit the number.

policies/documentation-policy.md
paulidale: I wanted it at the end as a note to indicate that this document was not all hard and fast rules.
It probably needs a bit more content.

slontis (commented):
looks ok to me.

policies/documentation-policy.md
```diff
@@ -0,0 +1,122 @@
...
+`INSTALL.md` file.
+
+## Chapter 9: Changes and news
+
+Significant modifications should be documented in the `CHANGES.md` file.
+
+Very significant features and changes should be documented in the `NEWS.md`
+file.
+
+In both cases, the added note should be short and to the point.
```
tmshort, line 108: This seems redundant, with `NEWS.md` being an abridged version of `CHANGES.md`.

I would suggest discarding `NEWS.md` and just reference `CHANGES.md`. "Very significant" features in `CHANGES.md` can be highlighted in some manner to identify them as such.

paulidale: I think both files have their use but I do see your point.
Possibly something for future discussion.

paulidale:
Vote for accepting the policy at the commit id 49f3b24e09fda7e32ac43aa221268b87d00ce233 has started.

paulidale:
Vote: [+1]

t-j-h:
Vote: [+1]

policies/documentation-policy.md
```diff
@@ -0,0 +1,122 @@
...
+- new OPENSSL_NO_DEPRECATED_ macros
+- new symbols generated automatically via `make update` (errors, objects, etc)
+
+## Chapter 3: Overviews, conventions, et al
+
+Where additional user facing information is required, it should be
+included in the `doc/man7` section.  This includes, but is not limited to:
+
+- descriptions of data structures and their usage
+- provider calls and call backs
```
vdukhovni, line 36:
```suggestion
- provider calls and callbacks
```

policies/documentation-policy.md
```diff
@@ -0,0 +1,122 @@
...
+These are functions, structures, globals and macros local to a specific
+C file or defined for a single directory as part of a local .h file.
+
+These should all be documented at the point of implementation.  Follow the
+rules and exceptions listed [above](#fsgm).  In some cases slightly more
+leniency with respect to _trivial_ can be tolerated.
+
+## Chapter 6: Code comments
+
+Comments in code should explain what the algorithm is doing and why.
```
vdukhovni, line 82: Here I'd more strongly emphasise the *why* over the *what*.  Design choices and rationale for the decisions made can be difficult to infer later. Block comments providing non-trivial  "backstory" details can be invaluable.
The *what* comments should be needed only when the implementation is subtly complex, a situation best avoided, but not always avoidable.

mattcaswell:
Vote: [+1]

policies/documentation-policy.md
kroeckx: I made the same comment 2 weeks ago. At that time it also started with "Comments in code should explain what the algorithm is doing". It seems the only effect of my comment is that "and why" got added to that sentence, which doesn't really address my problem with the text.

policies/documentation-policy.md
```diff
@@ -0,0 +1,122 @@
...
+The `make doc-nits` command should be run before submitting a pull
+request and any problems it locates must be addressed.
+
+## Chapter 11: Common sense
+
+Comments in the code are to improve readability and comprehension.
+Where the code is obvious, there is no need to include a comment.
+However, common sense applies.  Always err in favour of including more
+comments than less or none.  Code that you've just written that is
+_obvious_ will not necessarily be to someone else two years later.
```
kroeckx, line 121: It's also not because it's obvious now to you that it still will be 2 years from now.

policies/documentation-policy.md
```diff
@@ -0,0 +1,122 @@
+# OpenSSL documentation policy
+
+This document describes the code documentation and commenting
+requirements for the OpenSSL project.
+
+The project's documentation is all about making the libraries and tools
+more accessible to our users and making the code more maintainable.
+This policy applies to new submissions, existing code is below par and will
+be gradually improved.
```
kroeckx, line 9: I think our policy should not say that the existing code is below par. It also seems to say that existing documentation doesn't be fixed, it only applies to new code. But I expect it to say when making a non-trivial change to the code, the documentation should get checked and updated if needed.

policies/documentation-policy.md
```diff
@@ -0,0 +1,122 @@
...
+
+Where additional user facing information is required, it should be
+included in the `doc/man7` section.  This includes, but is not limited to:
+
+- descriptions of data structures and their usage
+- provider calls and call backs
+- algorithm descriptions and parameters
+- descriptions of architectural decisions
+
+## <a name="fsgm"></a>Chapter 4: Internal functions, structures, globals and macros
```
kroeckx, line 40: This is the only chapter having an `<a>`

kroeckx: Oh, there is a link to it.

policies/documentation-policy.md
vdukhovni: Agreed.  Here are some examples from Postfix:
https://github.com/vdukhovni/postfix/blob/master/postfix/src/tls/tls_dane.c#L458-L475
https://github.com/vdukhovni/postfix/blob/master/postfix/src/tls/tls_dane.c#L735-L739
https://github.com/vdukhovni/postfix/blob/master/postfix/src/tls/tls_client.c#L411-L425

t8m:
Vote: [+1]

beldmit:
Vote: [+1]

I am also thinking about commenting various flags/constants in-place, but not sure if it is worth separate mentioning.

kroeckx:
I think the text needs more work, so voting -1

slontis:
Vote: [+1]

mspncp:
Vote [+1]

mspncp (approved):

paulidale:
Merged

PR #21, opened 2022-01-26
code style: public functions should check their arguments

paulidale:
Public functions should include basic sanity checks of their arguments.

### DIFF
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -323,9 +323,9 @@ because it is a simple way to add valuable information for the reader.
 The name in the prototype declaration should match the name in the function
 definition.
 
-### Chapter 6.1: Checking function aguments
+### Chapter 6.1: Checking function arguments
 
-A _public_ function should verify that it's arguments are sensible.
+A _public_ function should verify that its arguments are sensible.
 This includes, but is not limited to, verifying that:
 - non-optional pointer arguments are not NULL and
 - numeric arguments are within expected ranges.
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -323,7 +323,16 @@ because it is a simple way to add valuable information for the reader.
 The name in the prototype declaration should match the name in the function
 definition.
 
-### Chapter 6.1: Extending existing functions
+### Chapter 6.1: Checking function aguments
+
+A _public_ function should verify that it's arguments are sensible.
+This includes, but is not limited to, verifying that:
+- non-optional pointer arguments are not NULL and
+- numeric arguments are within expected ranges.
+
+Where an argument is not sensible, an error should be returned.
+
+### Chapter 6.2: Extending existing functions
 
 From time to time it is necessary to extend an existing function. Typically this
 will mean adding additional arguments, but it may also include removal of some.
--- a/policies/stable-release-updates.md
+++ b/policies/stable-release-updates.md
@@ -50,6 +50,9 @@ A **bug fix** is **not**:
 - memory usage reduction
 - replacement implementation of algorithms
 - refactoring
+- addressing deviations from the [coding style policy].
+
+[coding style policy]: https://github.com/openssl/technical-policies/blob/master/policies/coding-style.md
 
 An **end-user documentation** is:
 
--- a/votes/vote-20220210-public-functions-check-args.txt
+++ b/votes/vote-20220210-public-functions-check-args.txt
@@ -1,0 +1,19 @@
+Topic: Public functions should check their arguments as of commit 1e37b5f.
+       This will become an official OTC policy.
+Proposed by: Pauli
+Issue link: https://github.com/openssl/technical-policies/pull/21
+Public: yes
+Opened: 2022-02-10
+Closed: 2022-02-17
+Accepted:  yes  (for: 8, against: 0, abstained: 1, not voted: 1)
+
+  Dmitry     [+1]
+  Matt       [+1]
+  Pauli      [+1]
+  Tim        [+1]
+  Richard    [  ]
+  Shane      [+1]
+  Tomas      [ 0]
+  Kurt       [+1]
+  Matthias   [+1]
+  Nicola     [+1]


policies/coding-style.md
```diff
@@ -323,7 +323,17 @@ because it is a simple way to add valuable information for the reader.
...
-### Chapter 6.1: Extending existing functions
+### Chapter 6.1: Checking function aguments
+
+A _public_ function should verify that it's arguments are sensible.
+This includes, but is not limited to, verifying that:
+- non-optional pointer arguments are not NULL;
+- enumerated arguments have a meaningful value and
+- numeric arguments are within expected ranges.
+
+Where an argument is not sensible, an error should be returned.
```
mattcaswell, line 334: Is failure to comply with the coding style a bug? I wonder if something needs to be said about this in the stable release updates policy. I am in favour of this change but am wary about opening the flood gates for all the issues and PRs relating to missing NULL checks in stable releases (or even master).

t8m: That's very good point and yeah, we should not treat failure to comply with coding style in an existing code a bug.

t8m: BTW a PR can modify multiple policy documents at once so if we want, we can modify the stable release updates policy in this PR as well. Although maybe it could be mentioned in the coding style policy instead. I am not sure which is the better place.

paulidale: The discussions around this were clear that fixing existing APIs wasn't going to happen quickly.

We also don't consider other coding style deviations to be bugs as far as I'm aware.

I'll add a not a bug note in the stable releases policy.

paulidale: Saying code style changes are not a bug should be in the stable releases policy.

votes/vote-20220125-revertn-aes-noasm.txt
```diff
@@ -0,0 +1,21 @@
+Topic: OTC approves the OMC request for an exception to the stable release
```
t8m, line 1: This is mistakenly added to the PR.

paulidale: Hmmmm, not sure where that is coming from.

paulidale: I'm only seeing two commits before I get to the master branch.  That's the one with the vote in it.
When I push, the third magically appears here.

t8m: Ah, the commit that added the vote was pushed to github.com mirror repo but not to the GHE primary repo.

t8m: So the 2592430 needs to be pushed to master on github.openssl.org and things should fix themselves.

policies/coding-style.md
```diff
@@ -323,7 +323,17 @@ because it is a simple way to add valuable information for the reader.
 The name in the prototype declaration should match the name in the function
 definition.
 
-### Chapter 6.1: Extending existing functions
+### Chapter 6.1: Checking function aguments
+
+A _public_ function should verify that it's arguments are sensible.
+This includes, but is not limited to, verifying that:
+- non-optional pointer arguments are not NULL;
+- enumerated arguments have a meaningful value and
```
kroeckx, line 331: I think this should also cover things like flags using bits, where it should reject unknown flags.

votes/vote-20220125-revertn-aes-noasm.txt
paulidale: Thanks.  I'd made a muddle of my clone of this repo and the primary repo was wrong...

policies/coding-style.md
paulidale: I'll add this specifically.  It could be assumed but best not.

t8m:
No idea why the https://github.com/openssl/technical-policies/pull/21/commits/fb7a6ea3bfa6d5b20be8718b5b6c89fd9a52ec49 is still listed here when it is already merged into the master branch.

policies/coding-style.md
t8m: I am worried that we are taking this argument checking too far.

paulidale:
Me either.  Thought I'd gone crazy.

policies/coding-style.md
t8m: Let's say that there is a new flag introduced to for example enable some security hardening in later releases. The older releases could work fine as well without enabling this hardening but since the flag passed in will be unknown the function call will fail.

mattcaswell: I agree with @t8m. Adding such checks could be unnecessarily constraining

kroeckx: I would even add that the *_ctrl() function should check that command is a known command, and the arguments (if any) might also have a limited amount of valid values. So I would in general say that if an argument should be check to be a known valid option.

Not checking that they are valid values almost always leads to problems. As far as I know, people have problems with 3.0 because we did not add such checks. Ignoring the unknown value can lead to security problems because they think your security hardening is available and have no way to check that it's not available and fall back to an other way to do it.

kroeckx:
I'm not sure that coding style is the correct document to put this in.

I consider not following this to be a bug, and don't consider adding such a check as a feature. But backports of such changes to stable branches are not a good idea.

policies/coding-style.md
paulidale: I've reverted this change.  I think @t8m hit a use case we want to work.

paulidale: Most of the ctrl functions do check that the command is known.
I agree, they should also check for valid arguments.  As should params.

kroeckx: I really don't understand the argument. If a function doesn't exist in an older version, we're happy with a (linker) error. If one of the control commands doesn't exist, we're happy with an error. But for the special case of flags, returning an error on an unknown flag is somehow problematic. Those things really are just the same thing, things we don't known about. Pretending we do known it and did and process the request is causing people problems.

paulidale:
> I'm not sure that coding style is the correct document to put this in.

Where then?  I can't think of a better place but am open to suggestions.

We are generally pretty good at mandating the coding style for contributions but there are always exceptions.  The exceptions lead me to thinking that it should not be a hard and fast policy (i.e. requires an OTC vote to not follow).  Some checks will be required, some can (mostly) safely be omitted.  We can't either mandate or relax here -- i.e. there needs to be some flexibility on a case by case basis.


> I consider not following this to be a bug, and don't consider adding such a check as a feature. But backports of such changes to stable branches are not a good idea.

It's kind of in the grey zone for me: if it's a bug, a back-port is fine; if it's a feature, back-porting is not allowed.  That's our defined policy.  A third option is to update the policy to allow these kind of changes to be neither a bug nor not.

paulidale:
Discussed at the OTC meeting and this is okay to go to vote after the fourteen day.
There was agreement that it should be in the coding style document.

paulidale:
Vote: [+1]

t8m:
Vote: [ 0]

policies/coding-style.md
```diff
@@ -323,7 +323,16 @@ because it is a simple way to add valuable information for the reader.
 The name in the prototype declaration should match the name in the function
 definition.
 
-### Chapter 6.1: Extending existing functions
+### Chapter 6.1: Checking function aguments
+
+A _public_ function should verify that it's arguments are sensible.
```
tom-cosgrove-arm, line 328: (minor) "it's" is short for "it is". The possessive is "its" (like "his", not like "John's")

paulidale: Argh!  I make this typo a lot :(

tom-cosgrove-arm: Yes, while I can't not notice the issue when I'm reading text, I find I frequently type both "its" and "it's" incorrectly!

mattcaswell:
Vote: [+1]

mspncp (commented):
Vote: [+1]

And I wouldn't mind if the two typos would get fixed on-the-fly while merging.  Thoughts, @openssl/otc ?

t8m:
> And I wouldn't mind if the two typos would get fixed on-the-fly while merging. Thoughts, @openssl/otc ?

Yeah, I feel like we should not try to be a bureaucratic machinery when a change is just about doing typo fixes.

paulidale:
I was thinking about pushing the referenced commit 1e37b5f0793e3e25d66408864d5acc5c32c664b4 and a second fix typos commit.  This meets the letter of the vote and leaves the trivial corrections clearly visible.

slontis:
Vote: [+1]

t-j-h:
Vote: [+1]

romen:
Vote: [+1]

beldmit:
Vote: [+1]

kroeckx:
Vote: +1

paulidale:
Vote closed.

Topic: Public functions should check their arguments as of commit 1e37b5f.
Proposed by: Pauli
Issue link: https://github.com/openssl/technical-policies/pull/21
Public: yes
Opened: 2022-02-10
Closed: 2022-02-17
Accepted:  yes  (for: 8, against: 0, abstained: 1, not voted: 1)

  Dmitry     [+1]
  Matt       [+1]
  Pauli      [+1]
  Tim        [+1]
  Richard    [  ]
  Shane      [+1]
  Tomas      [ 0]
  Kurt       [+1]
  Matthias   [+1]
  Nicola     [+1]

Merged after a rebase.

PR #24, opened 2022-02-04
Clarify the definition of stable release, and add definition of patch release

levitte:
The language used made it unclear if x.y.0 would be considered a
stable release.  Furthermore, the term "patch release" is used without
definition in this document.

### DIFF
--- a/policies/stable-release-updates.md
+++ b/policies/stable-release-updates.md
@@ -6,8 +6,10 @@ This policy covers allowed changes on stable release branches.
 Definitions
 -----------
 
-A **stable release** is a patch release from an existing supported minor
-release branch.
+A **stable release** is a series beginning with a major or minor release that
+is not a pre-release, and all its updates.
+
+A **patch release** is an update within a stable release.
 
 A **public interface** is any function, structure or macro declared in a public
 header file.
--- a/votes/vote-20220304-clarify-release-terminology.txt
+++ b/votes/vote-20220304-clarify-release-terminology.txt
@@ -1,0 +1,17 @@
+topic: Clarify the definition of stable release, and add definition of patch release
+Proposed by Richard
+Public: yes
+opened: 2022-03-04
+closed: 2022-03-23
+accepted:  yes  (for: 4, against: 0, abstained: 3, not voted: 3)
+
+  Dmitry     [  ]
+  Matt       [+1]
+  Pauli      [ 0]
+  Tim        [+1]
+  Richard    [+1]
+  Shane      [+0]
+  Tomas      [  ]
+  Kurt       [  ]
+  Matthias   [+1]
+  Nicola     [ 0]


policies/stable-release-updates.md
```diff
@@ -6,8 +6,9 @@ This policy covers allowed changes on stable release branches.
 Definitions
 -----------
 
-A **stable release** is a patch release from an existing supported minor
-release branch.
+A **stable release** is a major or minor release that is not a pre-release.
+
+A **patch release** is an update of a stable release.
```
mattcaswell, line 11: The wording implies that a patch release is not a stable release itself - just an update of a stable release. Could that be made clearer somehow?

It would also be useful to have standard terminology to refer to the whole series of releases rather than an individual patch level. I often find myself talking about the 3.0 release (where I mean any of 3.0.0, 3.0.1, 3.0.2 etc), and also the 3.0.1 release where I'm talking about a specific patch level.

mattcaswell (changes requested):

policies/stable-release-updates.md
t8m: Maybe call that a **stable release series**?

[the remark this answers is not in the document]
paulidale: I think x.y.0 should qualify as a patch release.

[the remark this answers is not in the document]
levitte: It's my understanding that it does.  If x.y.0 is a patch release, what is it a patch of?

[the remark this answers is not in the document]
mattcaswell: Its the "null" patch :-)

mattcaswell: Possibly. Although that's a bit of a mouthful. Maybe just "release series". Or even just "series".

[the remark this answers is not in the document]
levitte: Oh c'mon!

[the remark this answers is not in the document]
levitte: Also, "patch release" is mentioned on line 69 / 70

levitte: I've used "release update" for things like 1.1.1m or 3.0.1.  We have used "stable release" a bit vaguely, I'm never sure if we mean 3.0.0 or the whole series of 3.0.x...  I guess that's basically what @mattcaswell describes.  I frankly do not care what we end up calling it, but I would say that taking away "stable release" wouldn't be very good, since this whole document is a **stable release *policy***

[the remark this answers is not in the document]
mattcaswell: That line also talks about a "minor release", and the terms "major release" and "minor release" are used in your new wording without being defined.

So, IIUC, for any given release x.y.z that is not a pre-release:

If y and z are both 0 it is a major release.
If y is non-zero and z is 0 then it is a minor release
if z is non-zero it is a patch release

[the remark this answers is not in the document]
mattcaswell: Actually given that interpretation I'm not sure our use of "stable release" is quite right.

We later say:
"Only bug fixes, fixes for security issues, and end-user documentation
additions and fixes are allowed in stable releases."

And actually what we mean in that sentence is that they are only allowed in patch releases.

The stability that we are talking about in the "Stable Release Updates" policy is the stability of the whole series. So 1.1.1 is a stable release series as is 3.0.

mattcaswell: I could go with just defining "stable release" as being shorthand for "stable release series" which means the whole set 3.0.x. So you could talk about the 3.0 stable release or the 1.1.1 stable release.

levitte:
Text updated to clarify that a "stable release" is a major or minor release and its updates.

policies/stable-release-updates.md
```diff
@@ -6,8 +6,10 @@ This policy covers allowed changes on stable release branches.
 Definitions
 -----------
 
-A **stable release** is a patch release from an existing supported minor
-release branch.
+A **stable release** is a series including a major or minor release
+that is not a pre-release, and all its updates.
+
+A **patch release** is an update of a stable release.
```
mattcaswell, line 12: Hmm. The definition of "stable release" seems good now. But I'm not sure that "patch release" quite makes sense. Since the definition of "stable release" includes all the updates it doesn't make sense to say a patch release is an update of a stable release.

Possibly "A patch release is an update within a stable release series"?

levitte: Done

mattcaswell:
LGTM

policies/stable-release-updates.md
```diff
@@ -6,8 +6,10 @@ This policy covers allowed changes on stable release branches.
 Definitions
 -----------
 
-A **stable release** is a patch release from an existing supported minor
-release branch.
+A **stable release** is a series including a major or minor release
```
paulidale, line 9: _beginning with_ instead of _including_?

Something feels wrong here but I'm not able to put my finger on it.

levitte: Sure, seems like a good change

fariborz4:
Ji

levitte:
Ping

t8m (approved):

levitte:
Considering how long this has been lying around...  time to start the vote

t-j-h:
I don't think this actually improves things to pick a definition that isn't referring to the other definitions and simply uses different works which aren't actually defined ... in its present form I wouldn't support the update.

paulidale:
The usual process is to nominate a commit ID for the vote.....
It's not strictly written, but it's a good datum.

levitte:
> The usual process is to nominate a commit ID for the vote..... It's not strictly written, but it's a good datum.

Ok.  So, commit c741565d9f207b1b283946564b681755eed8b9b9

levitte:
Re @t-j-h's comment, I'm unsure how to respond precisely to that...  however, I'm starting to wonder if we're doing the right thing by having the final definition of terms spread around in diverse policies, or if the definitions (axioms, if you wish) should be in the glossary instead, and policies should refer to the glossary instead of the other way around.

This isn't a proposal to change this PR around, but rather a thought that would generate new PRs...

mattcaswell:
Vote: [+1]

t8m:
Vote: [+1]

paulidale:
I had similar thoughts about the glossary including the formal definitions.  In fact, I began with the idea that we'd have a definitions document and refer to it.  I changed things around to have a glossary instead fairly early on.  I figured the definitions were best in the policies that defined them -- they have the correct context after all.

However, I would be amenable to switching this around.  We'd need a definitions document not a glossary but that's a writing detail.  Plus, there will be large grey areas.....

paulidale:
Vote: [ 0]

mspncp:
> I'm starting to wonder if we're doing the right thing by having the final definition of terms spread around in diverse policies, or if the definitions (axioms, if you wish) should be in the glossary instead, and policies should refer to the glossary instead of the other way around.

Having the glossary as a central document where you can find all commonly used terms makes a lot of sense. IMO the purpose of the glossary is not to provide the „final definition“ of a term but rather a quick explanation and a link to the relevant definition (policy document, man page, etc) if necessary.

mspncp:
Vote: [+1]

romen:
Vote: [ 0]

slontis (commented):
Do you need to define what a pre-release is.

I am trying to think of an alternate wording for  'and all its updates.'  You basically want to say that patch updates can be applied to stable releases.

levitte:
> I am trying to think of an alternate wording for 'and all its updates.' You basically want to say that patch updates can be applied to stable releases.

Well, we have actually spoken of "stable release" as whatever the last update happens to be, i.e. as a synonym for the whole series / branch.  But granted, our use of "stable release" is dual, 'cause we have also used that to express the first release (i.e. 3.0.0 for the 3.0.x series)

slontis:
Vote: [+0]

policies/stable-release-updates.md
```diff
@@ -6,8 +6,10 @@ This policy covers allowed changes on stable release branches.
 Definitions
 -----------
 
-A **stable release** is a patch release from an existing supported minor
-release branch.
+A **stable release** is a series beginning with a major or minor release that
+is not a pre-release, and all its updates.
```
kroeckx, line 10: This sounds more like a stable branch, from which releases are made.

It's a stable release any release not marked as alpha, beta or release candidate?

levitte: > It's a stable release any release not marked as alpha, beta or release candidate?

For the moment, that's how we've used the term...

> This sounds more like a stable branch, from which releases are made.

That too.  But the way I understand our use of the word is that, for example, OpenSSL 3.0 is a stable release.  Does that just include 3.0.0, or does it include 3.0.x?  The way we've talked about it, the latter seems to apply.

levitte:
I'm closing this vote now.  It's been active for too long already

levitte:
Merged

b64ff64c5fa21d5c9a7c0c9f32d196b0bb2fcc13 Clarify the definition of stable release, and add definition of patch release
ff7436adf064ea2c886b242cc7da4621f720b22f Record the votes for "Clarify the definition of stable release, and add definition of patch release"

policies/stable-release-updates.md
mattcaswell: I definitely refer to 3.0 or 1.1.1 as a stable release. So the 3.0 stable release means any release from the series 3.0.x. There is some ambiguity in the term "release" here. Sometimes is means a specific release 3.0.1 or 3.0.2 for example. Sometimes it means the set of all releases from a stable branch, e.g. 3.0.x

kroeckx: For me 3.0 is a stable branch, the releases in it our 3.0.0, 3.0.1, ...

PR #32, opened 2022-02-28
Change date format of vote records from yyyymmdd to yyyy-mm-dd

mspncp:
This date format is better readable and coincides with the format
used inside the records and elsewhere.


paulidale (commented):
This also hits the OMC voting policy :(

mattcaswell:
It's a shame this wasn't raised when we discussed the recent vote to change the filename format #17.

I'm not sure its worth it since we'd also have to change the OMC policy to match - or live with the inconsistency between the 2 repos.

t8m:
-1 from me. IMO not worth it.

mspncp:
> It's a shame this wasn't raised when we discussed the recent vote to change the filename format https://github.com/openssl/technical-policies/pull/17.

Yeah, it‘s a pity. I was too occupied with my regular work to notice the inconsistency. Now it will have to bug me until the next update of the policy.

PR #39, opened 2022-04-01
coding-style.md: add various rules for writing C expressions

DDvO:
Add a first chunk of rules, taken from the list of proposed ones collected at https://github.com/openssl/openssl/issues/10725, 
focusing on how (not) to write C expressions. 

This PR significantly generalizes the former **Miscellaneous** section of [policies/coding-style.md](https://github.com/openssl/technical-policies/blob/master/policies/coding-style.md).

This PR only partially fixes https://github.com/openssl/openssl/issues/10725 - as @mattcaswell wrote in https://github.com/openssl/openssl/issues/10725#issuecomment-1085513555:
> I'd suggest any PR target only one or a handful of changes at a time.

### DIFF
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -593,39 +593,40 @@ used on and therefore should be avoided.
 
 ## Chapter 15: Expressions
 
-Avoid needless parentheses.<br/>
+Avoid needless parentheses.
 For example, do not write
 ```c
     if ((p == NULL) && (!f(((2 * x) + y) == (z++))))
- ```
+```
 but
 ```c
     if (p == NULL && !f(2 * x + y == z++)).
 ```
 
-For clarity, always put parentheses when mixing `&&` and `||` operations.<br/>
+For clarity, always put parentheses when mixing `&&` and `||` operations.
 For example,
- ```c
+```c
     if ((a && b) || c)
- ```
+```
 
-Put parentheses around macro bodies that can be used as expressions.<br/>
+Put parentheses around macro bodies that can be used as expressions.
+
 Put parentheses around uses of macro arguments
-(unless they are passed on as-is to a further macro or function).<br/>
+(unless they are passed on as-is to a further macro or function).
 For example,
- ```c
+```c
 #define SQUARE(a) ((a) * (a))
 ```
 
 In comparisons with constants (including `NULL`)
-place the constant on the right-hand side of the comparison operator.<br/>
+place the constant on the right-hand side of the comparison operator.
 For example,
 ```c
     while (i++ < 10 && p != NULL)
 ```
 
 Do not use implicit checks for
-numbers (not) being `0` or pointers (not) being `NULL`.<br/>
+numbers (not) being `0` or pointers (not) being `NULL`.
 For example, do not write
 ```c
     if (i)
@@ -644,7 +645,7 @@ but do this instead:
 If you need to break an expression into multiple lines,
 make the line break before an operator, not after.
 It is preferred that such a line break is made
-before as low priority an operator as possible.<br>
+before as low priority an operator as possible.
 Examples:
 
 * not this:
@@ -664,9 +665,8 @@ Examples:
       == PRETTY_DARN_LONG_MACRO_NAME)
   ```
 
-Whan appearing at the beginning of a line,
-the binary logical operators `&&` and `||`
-may get an extra indentation (+4 characters).<br/>
+When appearing at the beginning of a line,
+operators may get an extra indentation (+4 characters).
 For example,
 ```c
     if (long_condition_expression_1
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -255,7 +255,7 @@ variable that is used to hold a temporary value.
 
 If you are afraid that someone might mix up your local variable names,
 perhaps the function is too long;
-see the [chapter on functions](#user-content-chapter-6-functions).
+see the [chapter on functions](#chapter-6-functions).
 
 ## Chapter 5: Typedefs
 
@@ -623,7 +623,7 @@ For example,
 ```
 
 Regarding parentheses in macro definitions see the
-[chapter on macros](#user-content-chapter-9-macros-and-enums).
+[chapter on macros](#chapter-9-macros-and-enums).
 
 In comparisons with constants (including `NULL` and other constant macros)
 place the constant on the right-hand side of the comparison operator.
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -254,7 +254,8 @@ being mis-understood. Similarly, _tmp_ can be just about any type of
 variable that is used to hold a temporary value.
 
 If you are afraid that someone might mix up your local variable names,
-perhaps the function is too long; see Chapter 6.
+perhaps the function is too long;
+see the [chapter on functions](#user-content-chapter-6-functions).
 
 ## Chapter 5: Typedefs
 
@@ -469,16 +470,25 @@ Do not write macros that are l-values:
 
 This will cause problems if, e.g., FOO becomes an inline function.
 
-Be careful of precedence. Macros defining constants using expressions
-must enclose the expression in parentheses:
+Be careful of precedence.
+Macros defining an expression must enclose the expression in parentheses
+unless the expression is a literal or a function application:
 
 ```c
-    #define CONSTANT 0x4000
-    #define CONSTEXP (CONSTANT | 3)
+    #define SOME_LITERAL 0x4000
+    #define CONSTEXP (SOME_LITERAL | 3)
+    #define CONSTFUN foo(0, CONSTEXP)
 ```
 
-Beware of similar issues with macros using parameters. The [GNU cpp manual][8]
-deals with macros exhaustively.
+Beware of similar issues with macros using parameters.
+Put parentheses around uses of macro arguments
+unless they are passed on as-is to a further macro or function.
+For example,
+```c
+#define MACRO(a,b) ((a) * func(a, b))
+```
+
+The [GNU cpp manual][8] deals with macros exhaustively.
 
 ## Chapter 10: Allocating memory
 
@@ -512,7 +522,7 @@ subject to these rules. Generally they indicate failure by returning some
 out-of-range result. The simplest example is functions that return pointers;
 they return NULL to report failure.
 
-## Chapter 12:  Editor modelines
+## Chapter 12: Editor modelines
 
 Some editors can interpret configuration information embedded in source
 files, indicated with special markers. For example, emacs interprets
@@ -542,7 +552,7 @@ This includes markers for indentation and mode configuration. People may
 use their own custom mode, or may have some other magic method for making
 indentation work correctly.
 
-## Chapter 13:  Processor-specific code
+## Chapter 13: Processor-specific code
 
 In OpenSSL's case the only reason to resort to processor-specific code
 is for performance. As it still exists in a general platform-independent
@@ -584,7 +594,7 @@ less complicated than coding pure assembly, and it doesn't provide the
 same performance guarantee across different micro-architecture. Nor is
 it portable enough to meet our multi-platform support goals.
 
-## Chapter 14:  Portability
+## Chapter 14: Portability
 
 To maximise portability the version of C defined in ISO/IEC 9899:1990
 should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
@@ -593,7 +603,7 @@ used on and therefore should be avoided.
 
 ## Chapter 15: Expressions
 
-Avoid needless parentheses.
+Avoid needless parentheses as far as reasonable.
 For example, do not write
 ```c
     if ((p == NULL) && (!f(((2 * x) + y) == (z++))))
@@ -603,22 +613,19 @@ but
     if (p == NULL && !f(2 * x + y == z++)).
 ```
 
-For clarity, always put parentheses when mixing `&&` and `||` operations.
+For clarity, always put parentheses when mixing `&&` and `||` operations,
+comparison operators like `<=` and `==`, and bitwise operators.
 For example,
 ```c
     if ((a && b) || c)
+    if ((a <= b) == ((c >= d) != (e < f)))
+    x = (a & b) ^ (c | d)
 ```
 
-Put parentheses around macro bodies that can be used as expressions.
+Regarding parentheses in macro definitions see the
+[chapter on macros](#user-content-chapter-9-macros-and-enums).
 
-Put parentheses around uses of macro arguments
-(unless they are passed on as-is to a further macro or function).
-For example,
-```c
-#define SQUARE(a) ((a) * (a))
-```
-
-In comparisons with constants (including `NULL`)
+In comparisons with constants (including `NULL` and other constant macros)
 place the constant on the right-hand side of the comparison operator.
 For example,
 ```c
@@ -666,7 +673,7 @@ Examples:
   ```
 
 When appearing at the beginning of a line,
-operators may get an extra indentation (+4 characters).
+operators can, but do not have to, get an extra indentation (+ 4 characters).
 For example,
 ```c
     if (long_condition_expression_1
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -614,7 +614,7 @@ Put parentheses around uses of macro arguments
 (unless they are passed on as-is to a further macro or function).<br/>
 For example,
  ```c
-#define SQUARE(a) ((a) * (b))
+#define SQUARE(a) ((a) * (a))
 ```
 
 In comparisons with constants (including `NULL`)
@@ -641,9 +641,32 @@ but do this instead:
     if ((p = BN_new()) == NULL)
 ```
 
-Do not place binary logical operators `&&` and `||` at the end of a line
-but put them at the beginning of the following line.<br/>
-They may get an extra indentation (+4 characters).<br/>
+If you need to break an expression into multiple lines,
+make the line break before an operator, not after.
+It is preferred that such a line break is made
+before as low priority an operator as possible.<br>
+Examples:
+
+* not this:
+  ```c
+  if (somewhat_long_function_name(foo) == 1 && a_long_variable_name
+      == 2)
+  ```
+  but rather:
+  ```c
+  if (somewhat_long_function_name(foo) == 1
+      && a_long_variable_name == 2)
+  ```
+
+* This is, however, still ok:
+  ```c
+  if (this_thing->this_freakishly_super_long_name(somewhat_long_name, 3)
+      == PRETTY_DARN_LONG_MACRO_NAME)
+  ```
+
+Whan appearing at the beginning of a line,
+the binary logical operators `&&` and `||`
+may get an extra indentation (+4 characters).<br/>
 For example,
 ```c
     if (long_condition_expression_1
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -591,21 +591,66 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
 (also known as C99) is not supported on some platforms that OpenSSL is
 used on and therefore should be avoided.
 
-## Chapter 15: Miscellaneous
+## Chapter 15: Expressions
 
-Do not use `!` to check if a pointer is NULL, or to see if a str...cmp
-function found a match.  For example, these are wrong:
-
+Avoid needless parentheses.<br/>
+For example, do not write
 ```c
-    if (!(p = BN_new())) ...
-    if (!strcmp(a, "FOO")) ...
+    if ((p == NULL) && (!f(((2 * x) + y) == (z++))))
+ ```
+but
+```c
+    if (p == NULL && !f(2 * x + y == z++)).
 ```
 
-Do this instead:
+For clarity, always put parentheses when mixing `&&` and `||` operations.<br/>
+For example,
+ ```c
+    if ((a && b) || c)
+ ```
 
+Put parentheses around macro bodies that can be used as expressions.<br/>
+Put parentheses around uses of macro arguments
+(unless they are passed on as-is to a further macro or function).<br/>
+For example,
+ ```c
+#define SQUARE(a) ((a) * (b))
+```
+
+In comparisons with constants (including `NULL`)
+place the constant on the right-hand side of the comparison operator.<br/>
+For example,
 ```c
-    if ((p = BN_new()) == NULL)...
-    if (strcmp(a, "FOO") == 0) ...
+    while (i++ < 10 && p != NULL)
+```
+
+Do not use implicit checks for
+numbers (not) being `0` or pointers (not) being `NULL`.<br/>
+For example, do not write
+```c
+    if (i)
+    if (!(x & MASK))
+    if (!strcmp(a, "FOO"))
+    if (!(p = BN_new()))
+```
+but do this instead:
+```c
+    if (i != 0)
+    if ((x & MASK) == 0)
+    if (strcmp(a, "FOO") == 0)
+    if ((p = BN_new()) == NULL)
+```
+
+Do not place binary logical operators `&&` and `||` at the end of a line
+but put them at the beginning of the following line.<br/>
+They may get an extra indentation (+4 characters).<br/>
+For example,
+```c
+    if (long_condition_expression_1
+            && condition_expression_2) {
+        statement_1;
+        statement_2;
+    }
 ```
 
 ## Chapter 16: References


DDvO:
BTW, search engines still find https://www.openssl.org/policies/codingstyle.html
and https://github.com/openssl/web/blob/master/policies/codingstyle.txt -
would be good to add forward pointers there to the new document location
https://www.openssl.org/policies/technical/coding-style.html.

paulidale (approved):

paulidale:
As a policy change, this will require an OTC vote at least two weeks after the PR was first raised.

policies/coding-style.md
```diff
@@ -591,21 +591,66 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+but do this instead:
+```c
+    if (i != 0)
+    if ((x & MASK) == 0)
+    if (strcmp(a, "FOO") == 0)
+    if ((p = BN_new()) == NULL)
+```
 
+Do not place binary logical operators `&&` and `||` at the end of a line
+but put them at the beginning of the following line.<br/>
```
levitte, line 645: I would rephrase this sentence:

"If you need to break an expression into multiple lines, make the line break before an operator, not after.  It's preferred that such a line break is made before as low priority an operator as possible"

Examples:

-   not this:

    ``` C
    if (somewhat_long_function_name(foo) == 1 && a_long_variable
        == 2)
    ```

    but rather:

    ``` C
    if (somewhat_long_function_name(foo) == 1
        && a_long_variable == 2)
    ```

-   This is, however, still ok:

    ``` C
    if (this_thing->this_freakishly_super_long_name(somewhat_long_name, 3)
    == PRETTY_DARN_LONG_MACRO_NAME)
    ```

policies/coding-style.md
```diff
@@ -591,21 +591,66 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+They may get an extra indentation (+4 characters).<br/>
```
levitte, line 646: I would add an empty line above this one, to separate it from the expression line breaking.  It's not the same topic, even though somewhat related...

policies/coding-style.md
```diff
@@ -591,21 +591,66 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
 (also known as C99) is not supported on some platforms that OpenSSL is
 used on and therefore should be avoided.
 
-## Chapter 15: Miscellaneous
+## Chapter 15: Expressions
 
-Do not use `!` to check if a pointer is NULL, or to see if a str...cmp
-function found a match.  For example, these are wrong:
+Avoid needless parentheses.<br/>
+For example, do not write
```
zuiderkwast, line 597: Normally a new paragraph is used in text. A blank line achieves that. `<br>` is normally only used in poems, postal addresses and such.

In this case, maybe no new paragraph is needed though.

```suggestion
Avoid needless parentheses. For example, do not write
```

zuiderkwast (commented):
Nice. I was going to create this PR myself.

policies/coding-style.md
```diff
@@ -591,21 +591,66 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+ ```c
+    if ((a && b) || c)
+ ```
+
+Put parentheses around macro bodies that can be used as expressions.<br/>
+Put parentheses around uses of macro arguments
+(unless they are passed on as-is to a further macro or function).<br/>
+For example,
+ ```c
+#define SQUARE(a) ((a) * (b))
```
zuiderkwast, line 617:
```suggestion
#define SQUARE(a) ((a) * (a))
```

policies/coding-style.md
DDvO: I've added the paragraph break and shortened these two sentences as part of them is now covered by the above generalized rule.

policies/coding-style.md
DDvO: Right. I tend to add explicit newlines when I feel this make the text more clear/readable.

policies/coding-style.md
DDvO: Oops :)

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
-## Chapter 15: Miscellaneous
+## Chapter 15: Expressions
 
-Do not use `!` to check if a pointer is NULL, or to see if a str...cmp
-function found a match.  For example, these are wrong:
+Avoid needless parentheses.<br/>
+For example, do not write
+```c
+    if ((p == NULL) && (!f(((2 * x) + y) == (z++))))
+ ```
```
levitte, line 600: These ` ``` ` are badly aligned

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+but
+```c
+    if (p == NULL && !f(2 * x + y == z++)).
+```
+
+For clarity, always put parentheses when mixing `&&` and `||` operations.<br/>
+For example,
+ ```c
+    if ((a && b) || c)
+ ```
```
levitte, line 610: Both ` ```c ` and ` ``` ` are badly aligned

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+
+Put parentheses around macro bodies that can be used as expressions.<br/>
+Put parentheses around uses of macro arguments
+(unless they are passed on as-is to a further macro or function).<br/>
+For example,
+ ```c
```
levitte, line 616: ` ```c ` is badly aligned

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+
+* This is, however, still ok:
+  ```c
+  if (this_thing->this_freakishly_super_long_name(somewhat_long_name, 3)
+      == PRETTY_DARN_LONG_MACRO_NAME)
+  ```
+
+Whan appearing at the beginning of a line,
+the binary logical operators `&&` and `||`
+may get an extra indentation (+4 characters).<br/>
```
levitte, line 669: Is there a reason why those operators are singled out?

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+      && a_long_variable_name == 2)
+  ```
+
+* This is, however, still ok:
+  ```c
+  if (this_thing->this_freakishly_super_long_name(somewhat_long_name, 3)
+      == PRETTY_DARN_LONG_MACRO_NAME)
+  ```
+
+Whan appearing at the beginning of a line,
```
levitte, line 667: "When"

policies/coding-style.md
zuiderkwast: ok. I think it looks like a typographic error mostly. In a paper or book, a new paragraph usually looks like a newline with some indentation on the next line, while on the web it's almost always vertical spacing. It's still a new paragraph in both cases though, just different rendering.

levitte:
Re `<br/>`, I agree that if they are meant as paragraph separators, then replace them with an empty line.  Paragraph style should NOT be a markdown matter, but a rendering (stylesheet) matter.

t8m (changes requested):
Please drop the `<br/>` marks. Replace with blank lines where appropriate.

policies/coding-style.md
DDvO: All right, removed all the `<br/>`.

policies/coding-style.md
DDvO: Oops, fixed all these alignments.

zuiderkwast (approved):

DDvO:
> As a policy change, this will require an OTC vote at least two weeks after the PR was first raised.

Would it be adequate to start with such an OTC vote now, or after two weeks of waiting time?

t8m:
It has to wait.

policies/coding-style.md
```diff
@@ -641,9 +641,32 @@ but do this instead:
     if ((p = BN_new()) == NULL)
 ```
 
-Do not place binary logical operators `&&` and `||` at the end of a line
-but put them at the beginning of the following line.<br/>
-They may get an extra indentation (+4 characters).<br/>
+If you need to break an expression into multiple lines,
+make the line break before an operator, not after.
+It is preferred that such a line break is made
+before as low priority an operator as possible.<br>
```
levitte: Remove `<br>` and add an empty line before `Example:`

policies/coding-style.md
zuiderkwast: It's generalized in the last commit:

````Markdown
When appearing at the beginning of a line,
operators may get an extra indentation (+4 characters).
For example,
```c
    if (long_condition_expression_1
            && condition_expression_2) {
        statement_1;
        statement_2;
    }
```
````

I think this comment can be resolved.

File: policies/coding-style.md (hunk 10 above)
Comment (levitte) [reply policies/coding-style.md]: Ah, must have been because I followed the link to a specific fixup commit, not realising that wasn't the last one.

File: policies/coding-style.md (hunk 8 above)
Comment (levitte) [reply policies/coding-style.md]: Yup

**Review by levitte — APPROVED** [review]

File: policies/coding-style.md [hunk 11]
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+
+Put parentheses around macro bodies that can be used as expressions.
 
+Put parentheses around uses of macro arguments
+(unless they are passed on as-is to a further macro or function).
+For example,
 ```c
-    if (!(p = BN_new())) ...
-    if (!strcmp(a, "FOO")) ...
+#define SQUARE(a) ((a) * (a))
```
slontis, line 618: Not sure if the macro section would be a better place for these?

DDvO: Good point that there is an overlap with the macro section.
Yet maybe some redundancy is even good in the sense that some reader may only have a look at either of these two place when looking for guidance for macro expressions?
And pragmatically, any further change to this PR would invalidate the approvals already given.

zuiderkwast: I agree, this is about macros rather than about expressions.

> any further change to this PR would invalidate the approvals already given.

This is an example of where the rules (two weeks quarantine, vote, etc.) are not helping but preventing improvements.

If there's a new rule about trivial changes that I've heard mentioning, then perhaps in the future, moving around and reformulating information (without changing the essence of it) can be considered trivial and done without the full process... (I hope so.)

DDvO: Here is a suggestion for a pragmatic way out in this case:
I anyway plan to provide further chunks, extending this document with further items of https://github.com/openssl/openssl/issues/10725#issue-543528689.
As far as requested, I can simply sort out overlaps between the two sections with the next PR.

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+For example, do not write
+```c
+    if ((p == NULL) && (!f(((2 * x) + y) == (z++))))
+```
+but
+```c
+    if (p == NULL && !f(2 * x + y == z++)).
+```
+
+For clarity, always put parentheses when mixing `&&` and `||` operations.
```
paulidale, line 606: I suspect we will want a whole lot more exceptions.  C has lots of operators and almost nobody learns the precedence order.  Perhaps this is a _common sense_ rule, reduce parenthesis as far as is reasonable but no further?

E.g. `a <= b == c >= d + 1` is a perfectly well defined C expression with minimal parenthesis but a lot of folks won't instantly know what it's doing.

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
...
+  ```
+
+* This is, however, still ok:
+  ```c
+  if (this_thing->this_freakishly_super_long_name(somewhat_long_name, 3)
+      == PRETTY_DARN_LONG_MACRO_NAME)
+  ```
+
+When appearing at the beginning of a line,
+operators may get an extra indentation (+4 characters).
```
paulidale, line 669: s/may/can, but do not have to,/

DDvO: Done

policies/coding-style.md
DDvO: I've extended the text and the examples this way, also regarding the so-called bitwise operators `&`, `^`, and `|`.

policies/coding-style.md
```diff
@@ -254,7 +254,8 @@ being mis-understood. Similarly, _tmp_ can be just about any type of
 variable that is used to hold a temporary value.
 
 If you are afraid that someone might mix up your local variable names,
-perhaps the function is too long; see Chapter 6.
+perhaps the function is too long;
+see the [chapter on functions](#user-content-chapter-6-functions).
```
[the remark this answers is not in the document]
DDvO: Done, and also for the other cross reference.

t8m (approved):

t8m:
> Can we meanwhile proceed to the mentioned OTC vote?

We need to discuss the proposal within OTC. I'll try to bring it up tomorrow if there is a time.

DDvO:
I never saw the blue 'reviewed' label which god added recently, interesting.
Strange that at least my browser does not list it in the above right corner, where there is already the red 'policy change' label.

policies/coding-style.md
t-j-h: The whole approach of preventing "extra" parentheses is something I think should be removed completely - as mistakes do happen and parentheses do significantly improve readability.

policies/coding-style.md
```diff
@@ -591,21 +591,89 @@ should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
 (also known as C99) is not supported on some platforms that OpenSSL is
 used on and therefore should be avoided.
 
-## Chapter 15: Miscellaneous
+## Chapter 15: Expressions
 
-Do not use `!` to check if a pointer is NULL, or to see if a str...cmp
-function found a match.  For example, these are wrong:
+Avoid needless parentheses.
```
t-j-h: As in I would remove the "Avoid needless parentheses" entirely ...

t8m: Why? Maybe make it a little less imperative? `Needless parentheses should be avoided.`

t8m: Actually the current wording is `Avoid needless parentheses as far as reasonable.` And IMO that's OK.

policies/coding-style.md
t8m: @t-j-h I wholeheartedly disagree. For me a code with needless parentheses is much less readable actually. I always have to count the parentheses and so on to understand what is going on there. The example in this PR is a perfect one for me. The condition with the extra parentheses looks like parentheses hell to me where the one without is perfectly readable.

beldmit: In most real life cases I'd agree with @t-j-h. I never remember the operators priority :(

DDvO: Well, what to consider a reasonable amount of extra parentheses is a matter of both taste and fluency with C.
This could be debated ad nauseam.

zuiderkwast: If this ever becomes a problem in practice (in a code review) just update the coding style again when that happens?

t8m:
Vote for accepting the coding style changes at f2ae669428dbc29b9d799ebfde2fea0b0ef45746 has started.

t8m:
Vote: [+1]

slontis:
Vote: [+1]

paulidale:
Vote: [+1]

paulidale (approved):

beldmit:
Vote: [+1]

t-j-h:
Vote: [+1]

romen:
Vote: [+0]

policies/coding-style.md
```diff
@@ -584,28 +594,93 @@ less complicated than coding pure assembly, and it doesn't provide the
...
+```c
+    if ((p == NULL) && (!f(((2 * x) + y) == (z++))))
+```
+but
+```c
+    if (p == NULL && !f(2 * x + y == z++)).
+```
 
+For clarity, always put parentheses when mixing `&&` and `||` operations,
+comparison operators like `<=` and `==`, and bitwise operators.
```
levitte, line 617: This one surprised me 😕

levitte: So now, `(a == 1 || b == 2)` is badly seen?  We have written that sort of expression all over the place until now...

DDvO: This recent addition was motivated by https://github.com/openssl/technical-policies/pull/39#discussion_r846933175.

levitte:
Vote: [-0]

policies/coding-style.md
paulidale: That comment was intended as the kind of mess C operators can create.  I don't think it's a great example to make a rule from, but I can live with it.  I have seen code along the lines of `a < b != c < d`.

I think we need some flexibility in interpretation.  Minimal parenthesis is an ideal but needs to be tempered but comprehension.

levitte: > I think we need some flexibility in interpretation

I agree on principle, but have a hard time seeing the flexibility in "always"

mspncp: I can only agree with this sentence if the **and** is taken literally:  parentheses are only required if you are mixing operators from all three categories

- (A) comparisons [6,7] 
- (B) binary bitwise boolean operators (8,9,10)
- (C) binary logical boolean operators  (11,12)

A simple expression like `(a == 1 || b == 2 && c == 3)` which only combines comparisons (A) using logical booleans (C) should not require parentheses. (And just in case: whoever does not know that AND binds stronger than OR in C should be ashamed an reread his copy of K&R.)

grafik

(see https://en.cppreference.com/w/c/language/operator_precedence)

mspncp:
Vote: [0]

policies/coding-style.md
t8m: I've understood it (and I think it was meant to be) as mixing operators in the same category only. I.E. this does not apply to `a == b && c > d`. But applies to `a && b || c ` which should be written as `(a && b) || c` even if the former is equivalent according to C rules.

t8m: > (And just in case: whoever does not know that AND binds stronger than OR in C should be ashamed an reread his copy of K&R.)

It is not a problem of shame. It is a problem of easy readability from top of your head - I.E. so you do not have to think about it.

t8m: > So now, `(a == 1 || b == 2)` is badly seen? We have written that sort of expression all over the place until now...

Nope. This should be OK with the policy as is, it is obvious from the other examples in the coding style policy. Yeah, the sentence might be clarified under the minor policy edits policy, once we have it approved. :grin:

mspncp: Personally, I really like K&R's decision to give the AND operator precedence over the OR operator and dislike putting unnecessary parentheses there, but I wouldn't be dogmatic about it. My K&R potshot was not meant seriously.

mspncp: Personally, I really like K&R's decision to give the AND operator precedence over the OR operator and dislike putting unnecessary parentheses there, but I wouldn't be dogmatic about it. My K&R potshot was not meant seriously.

DDvO: I would have hoped that the examples make clear how to understand the sentence in question.
Adding the sentence suggested by @t8m would certainly help clarifying the intended meaning.
Should I do although this might invalidate the approvals and the votes already cast?

t8m: Please keep it as is. I think it can be added after the minor policy edits is approved.

mattcaswell: > I think it can be added after the minor policy edits is approved.

It would probably change the meaning...

mattcaswell: The way I read it the sentence above is completely contradicting the immediately previous example.

policies/coding-style.md
```diff
@@ -584,28 +594,93 @@ less complicated than coding pure assembly, and it doesn't provide the
...
+
+In comparisons with constants (including `NULL` and other constant macros)
+place the constant on the right-hand side of the comparison operator.
+For example,
+```c
+    while (i++ < 10 && p != NULL)
+```
 
+Do not use implicit checks for
+numbers (not) being `0` or pointers (not) being `NULL`.
```
mattcaswell, line 636: Where the number is intended to be a boolean value I think this is ok and is *very* common in the codebase (in particular when checking function return values)m e.g. `if (!EVP_DecryptUpdate(...))`

mattcaswell:
Vote: [-1]

I broadly support this, but there are a couple of details I'm not happy with.

policies/coding-style.md
DDvO: > > I think it can be added after the minor policy edits is approved.
> 
> It would probably change the meaning...

No, the sentence @t8m proposed to add later:
> I.E., do not depend on `&&` having higher precedence than `||`, `<=` having higher precedence than `==`, and `&` having higher precedence than `|` or `^`.

would just clarify the sentence before:

> For clarity, always put parentheses when mixing `&&` and `||` operations, comparison operators like `<=` and `==,` and bitwise operators.

policies/coding-style.md
DDvO: Sure, and therefore the rule is stated just about *numbers* (integers etc.), not about Boolean values.
Maybe this should be made more clear.

DDvO: @mattcaswell it would have been good to receive such comments regarding clarifications earlier,
not after most people have already voted.

policies/coding-style.md
levitte: > I've understood it (and I think it was meant to be) as mixing operators in the same category only. I.E. this does not apply to `a == b && c > d`. But applies to `a && b || c ` which should be written as `(a && b) || c` even if the former is equivalent according to C rules.

Upon re-reading it, I can understand the sentence like that. But it's not perfectly clear. I suspect that we might have disputes later on depending on interpretation.

Ah well

paulidale:
With the discussion, I'm changing my vote because I think we need additional clarifications:

Vote: [-1]

levitte:
Changing my vote too for the current text.

Vote: [-1]

policies/coding-style.md
levitte, line 617: Looking at the sentence again, it wouldn't be very hard to make it clearer:

```suggestion
For clarity, always put parentheses when mixing operators of the same category,
such as `&&` and `||` operations, or comparison operators like `<=` and `==`, 
or bitwise operators.
```

policies/coding-style.md
DDvO: To clarify this, we could add (after the given examples) for instance:

Note that Boolean values shall be used directly as usual, e.g.,
`if (check(x) && !success(y))`

policies/coding-style.md
DDvO: Hmm, when re-reading the above suggestion, which does include an improvement, I do not like the "category" and "such as" very much.
For instance, this could also be understood that one should put parentheses when mixing `*` and `+`, which is not intended.
So I propose a further refinement as follows:

For clarity, always put parentheses when mixing the Boolean `&&` and `||` operations,
when mixing comparison operators like `<=` and `==`, and when when mixing bitwise operators.

levitte: > For clarity, always put parentheses when mixing the Boolean `&&` and `||` operations, when mixing comparison operators like `<=` and `==`, and when when mixing bitwise operators.

"Boolean" should be "Logical", to match C parlance.
The last "and" should probably be "or"
Drop the double "when"

With those mods, I approve

levitte: Oh, and "operations" should be "operators"

DDvO: Yes, and I suggest including in the sentence also examples for the bitwise operators.
So:

For clarity, always put parentheses when mixing the logical `&&` and `||` operators,
mixing comparison operators like `<=` and `==,` or mixing bitwise operators like  `&` and `|` .

paulidale: Any changes will require a new vote -- too many have voted to change the wording now.
We don't have a process to abort a vote without it failing.

levitte: > Any changes will require a new vote -- too many have voted to change the wording now. We don't have a process to abort a vote without it failing.

Does that mean that a new PR is needed, or can we cast a new vote in the same PR?
After all, our votes are tied to a specific *commit*, is it not?

DDvO: In my view a new PR would be needlessly inefficient.
How about doing the above clarification and the one I proposed in https://github.com/openssl/technical-policies/pull/39#discussion_r864516234
and then restart the voting on the new state?

(As apparently needed according to the policy, the current vote will need to be made fail first.)

mspncp:
Vote: [-1]

I'm not opposed against this PR in general, but I see some need for more discussion as well.

policies/coding-style.md
mspncp: @DDvO I agree with Richard: IMHO a PR which has been voted down should be closed. In cases like this one, were it just failed because changes were requested, it can immediately be replaced by a new PR from the same branch (or with some suffix like `coding-style-expressions-r2`, as you prefer).

Having multiple votes within the same PR would make the thread history too confusing. But at this moment, that's just my personal opinion Your thought, @openssl/otc ?

mspncp: Make it an **or**, not an **and**

> For clarity, always put parentheses when mixing the Boolean && and || operations,
> when mixing comparison operators like <= and ==, **or** when when mixing bitwise operators.

mspncp: > Any changes will require a new vote -- too many have voted to change the wording now. We don't have a process to abort a vote without it failing.

I don't think there is a need for different failure modes. If the vote fails, it fails. But if the OTC members indicate that a revised version would be accepted, then discussion can be immediately restarted on a new PR.

DDvO: > @DDvO I agree with Richard: IMHO a PR which has been voted down should be closed. In cases like this one, were it just failed because changes were requested, it can immediately be replaced by a new PR from the same branch (or with some suffix like `coding-style-expressions-r2`, as you prefer).

True, at least when using the same branch, a new PR is very easily set up.
Yet a pointer to the previous PR should be given such that original discussion can be found quickly.

> Having multiple votes within the same PR would make the thread history too confusing. But at this moment, that's just my personal opinion Your thoughts, @openssl/otc ?

That's a good point in favor of a new PR.

DDvO:
I'm going to set up a new PR with the two improvements proposed above this evening (European time).
Are there any other proposals for (further) changes?

policies/coding-style.md
paulidale: The vote includes the hash.  I don't see a big problem having a second vote here.
Another alternative would be an issue for the vote, keeping this PR open.

t8m:
Changing my vote so I can immediately close the vote.

Vote: [-1]

mspncp:
> I'm going to set up a new PR with the two improvements proposed above this evening (European time). Are there any other proposals for (further) changes?

There will be a review period as usual. And this time we promise to review _before_ voting 😉.

paulidale:
Another concern with a new PR is "does the waiting period reset"?

t8m:
> Another concern with a new PR is "does the waiting period reset"?

There is no escape route for that. So yes.

mspncp:
Yes it does, at least until we change the policy. (We could decide to have a shorter waiting period for revised PRs in the future.)

t8m:
Vote is closed. The change is rejected.

```
Topic: coding-style.md: add various rules for writing C expressions
       at commit f2ae669
Proposed by: Tomas
Issue link: https://github.com/openssl/technical-policies/pull/39
Public: yes
Opened: 2022-05-03
Closed: 2022-05-04
Accepted: no   (for: 3, against: 5, abstained: 1, not voted: 1)

   Dmitry     [+1]
   Matt       [-1]
   Pauli      [-1]
   Tim        [+1]
   Richard    [-1]
   Shane      [+1]
   Tomas      [-1]
   Kurt       [  ]
   Matthias   [-1]
   Nicola     [+0]
```

t8m:
Closing. Further discussion should be in the new pr once it is open.

policies/coding-style.md
t8m: The current process IMO does not allow for second vote. If anyone wants to propose a change to it no problem. I suppose we would need to change it to always use a separate issue for the votes.

mspncp: @t8m this is actually a very good idea: we shouldn't carry out the vote in the PR, but always create a dedicated Issue for a vote, with a standardized subject line (`VOTE: <subject>`). In this way, we could decouple the PR from the vote. Nevertheless, we need to amend the policies to regulate things like a shortened waiting period, if desired.

kroeckx:
So I'm voting -1 too

PR #48, opened 2022-05-04
coding-style.md: add various rules on names

DDvO:
Add a 3rd chunk of rules, taken from the list of proposed ones collected at https://github.com/openssl/openssl/issues/10725, 
focusing on names.

On this occasion clean up redundancy in the Naming chapter and re-order the rules.

### DIFF
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -240,9 +240,11 @@ Local variable names should be short, and to the point. If you have
 some random integer loop counter, it should probably be called `i` or `j`.
 Apart from that, avoid single-letter names,
 in particular when they can be visually confusing, such as `I` and `O`.
-Calling a varable `loop_counter` is non-productive, if there is no chance of it
-being mis-understood. Similarly, `tmp` can be just about any type of
-variable that is used to hold a temporary value.
+
+Calling a variable `loop_counter` is non-productive,
+if there is no chance of it being mis-understood.
+Use simple variable names like `tmp` and `name`
+as long as they are non-ambiguous in the given context.
 
 If you are afraid that someone might mix up your local variable names,
 perhaps the function is too long; see Chapter 6.
@@ -268,6 +270,10 @@ Use uppercase prefix like `EVP_` or `OSSL_CMP_` for public (API) symbols.
 Do not encode the type into a name (so-called Hungarian notation).
 
 Align names to terms and wording used in standards and RFCs.
+
+Avoid mixed-case unless needed by other rules.
+Especially never use `FirstCharacterUpperCase`.
+For instance, use `EVP_PKEY_do_something` rather than `EVP_DigestDoSomething`.
 
 Except when otherwise required, avoid mixed-case names.
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -244,8 +244,6 @@ such as `I` and `O`.
 Avoid other single-letter names unless they are telling in the given context.
 For instance, `m` for modulus and `s` for SSL pointers are fine.
 
-Calling a variable `loop_counter` is non-productive,
-if there is no chance of it being mis-understood.
 Use simple variable names like `tmp` and `name`
 as long as they are non-ambiguous in the given context.
 
@@ -270,7 +268,7 @@ unless they are `static` (i.e., local to the source file).
 
 Use uppercase prefix like `EVP_` or `OSSL_CMP_` for public (API) symbols.
 
-Do not encode the type into a name (so-called Hungarian notation).
+Do not encode the type into a name (so-called Hungarian notation, e.g., `int iAge`).
 
 Align names to terms and wording used in standards and RFCs.
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -234,27 +234,40 @@ changing their context lines.
 
 ## Chapter 4: Naming
 
-C is a Spartan language, and so should your naming be. Do not use long
-names like _ThisVariableIsATemporaryCounter_. Use a name like _tmp_, which
-is much easier to write, and not more difficult to understand.
+C is a Spartan language, and so should your naming be.
 
-Except when otherwise required, avoid mixed-case names.
+Local variable names should be short, and to the point. If you have
+some random integer loop counter, it should probably be called `i` or `j`.
+Apart from that, avoid single-letter names,
+in particular when they can be visually confusing, such as `I` and `O`.
+Calling a varable `loop_counter` is non-productive, if there is no chance of it
+being mis-understood. Similarly, `tmp` can be just about any type of
+variable that is used to hold a temporary value.
 
-Do not encode the type into a name (so-called Hungarian notation).
+If you are afraid that someone might mix up your local variable names,
+perhaps the function is too long; see Chapter 6.
 
 Global variables (to be used only if you REALLY need them) need to
 have descriptive names, as do global functions. If you have a function
 that counts the number of active users, you should call that
 _count_active_users()_ or similar, you should NOT call it _cntusr()_.
 
-Local variable names should be short, and to the point. If you have
-some random integer loop counter, it should probably be called _i_.
-Calling it _loop_counter_ is non-productive, if there is no chance of it
-being mis-understood. Similarly, _tmp_ can be just about any type of
-variable that is used to hold a temporary value.
+For getter and setter functions use names containing `get0_` or `get1_`
+(rather than `get_`) and `set0_` or `set1_` (rather than `set_`)
+and `push0_` or `push1_` (rather than `push_`)
+to point out whether the value parameter has aliasing or copying semantics.
 
-If you are afraid that someone might mix up your local variable names,
-perhaps the function is too long; see Chapter 6.
+Use lowercase prefix like `ossl_` for internal symbols.
+
+Use uppercase prefix like `EVP_` or `OSSL_CMP_` for public (API) symbols.
+
+Do not encode the type into a name (so-called Hungarian notation).
+
+Align names to terms and wording used in standards and RFCs.
+
+Except when otherwise required, avoid mixed-case names.
+
+Make sure that names do not contain spelling errors.
 
 ## Chapter 5: Typedefs
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -252,12 +252,15 @@ have descriptive names, as do global functions. If you have a function
 that counts the number of active users, you should call that
 _count_active_users()_ or similar, you should NOT call it _cntusr()_.
 
-For getter and setter functions use names containing `get0_` or `get1_`
+For getter functions returning pointers
+and functions setting a pointer given as a parameter,
+use names containing `get0_` or `get1_`
 (rather than `get_`) and `set0_` or `set1_` (rather than `set_`)
 and `push0_` or `push1_` (rather than `push_`)
 to point out whether the value parameter has aliasing or copying semantics.
 
-Use lowercase prefix like `ossl_` for internal symbols.
+Use lowercase prefix like `ossl_` for internal symbols
+unless they are `static` (i.e., local to the source file).
 
 Use uppercase prefix like `EVP_` or `OSSL_CMP_` for public (API) symbols.
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -238,8 +238,11 @@ C is a Spartan language, and so should your naming be.
 
 Local variable names should be short, and to the point. If you have
 some random integer loop counter, it should probably be called `i` or `j`.
-Apart from that, avoid single-letter names,
-in particular when they can be visually confusing, such as `I` and `O`.
+
+Avoid single-letter names when they can be visually confusing,
+such as `I` and `O`.
+Avoid other single-letter names unless they are telling in the given context.
+For instance, `m` for modulus and `s` for SSL pointers are fine.
 
 Calling a variable `loop_counter` is non-productive,
 if there is no chance of it being mis-understood.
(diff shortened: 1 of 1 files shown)


policies/coding-style.md
```diff
@@ -234,27 +234,40 @@ changing their context lines.
...
 
-Local variable names should be short, and to the point. If you have
-some random integer loop counter, it should probably be called _i_.
-Calling it _loop_counter_ is non-productive, if there is no chance of it
-being mis-understood. Similarly, _tmp_ can be just about any type of
-variable that is used to hold a temporary value.
+For getter and setter functions use names containing `get0_` or `get1_`
+(rather than `get_`) and `set0_` or `set1_` (rather than `set_`)
+and `push0_` or `push1_` (rather than `push_`)
+to point out whether the value parameter has aliasing or copying semantics.
```
levitte, line 258: This is only when the value argument or return value are pointers.  For any other argument, the 0 / 1 semantics is irrelevant.

policies/coding-style.md
```diff
@@ -234,27 +234,40 @@ changing their context lines.
...
 
-If you are afraid that someone might mix up your local variable names,
-perhaps the function is too long; see Chapter 6.
+Use lowercase prefix like `ossl_` for internal symbols.
```
levitte, line 260: Unless they are static in the .c file.  In that case, the prefix is irrelevant...  and it might actually be useful not to have the prefix, to denote functions that are meant to be source file private.

policies/coding-style.md
DDvO: Right - I've added a respective restriction to the sentence.

policies/coding-style.md
DDvO: Yep, I've added a respective restriction to this sentence, too.

policies/coding-style.md
DDvO: I've just further improved the wording of this rule.

policies/coding-style.md
```diff
@@ -234,27 +234,44 @@ changing their context lines.
...
 
-C is a Spartan language, and so should your naming be. Do not use long
-names like _ThisVariableIsATemporaryCounter_. Use a name like _tmp_, which
-is much easier to write, and not more difficult to understand.
+C is a Spartan language, and so should your naming be.
 
-Except when otherwise required, avoid mixed-case names.
+Local variable names should be short, and to the point. If you have
+some random integer loop counter, it should probably be called `i` or `j`.
+Apart from that, avoid single-letter names,
```
t8m, line 241: This is too strict! There might be cases where even a single letter is descriptive enough. Especially if the function is very short. I suggest: `Avoid single-letter names when they can be visually confusing, such as I and O.`

policies/coding-style.md
```diff
@@ -234,27 +234,44 @@ changing their context lines.
...
+in particular when they can be visually confusing, such as `I` and `O`.
+Calling a varable `loop_counter` is non-productive, if there is no chance of it
+being mis-understood. Similarly, `tmp` can be just about any type of
+variable that is used to hold a temporary value.
```
t8m, line 245: I do not understand the purpose of the sentence with `tmp` - is that recommending to use `tmp` or against that?

policies/coding-style.md
```diff
@@ -234,27 +234,44 @@ changing their context lines.
...
+Use lowercase prefix like `ossl_` for internal symbols
+unless they are `static` (i.e., local to the source file).
+
+Use uppercase prefix like `EVP_` or `OSSL_CMP_` for public (API) symbols.
+
+Do not encode the type into a name (so-called Hungarian notation).
+
+Align names to terms and wording used in standards and RFCs.
+
+Except when otherwise required, avoid mixed-case names.
```
t8m, line 272: I am afraid this can be misinterpreted. Perhaps just banning FirstCharacterUpperCase would be sufficient?

```
Avoid mixed-case unless needed by other rules. Especially never use `FirstCharacterUpperCase`.

Good example: `EVP_PKEY_do_something`

Bad example: `EVP_DigestDoSomething'
```

t8m (changes requested):

policies/coding-style.md
DDvO: I don't know, either.
This sentence is part of the original text - it appears 'new' just because I re-arranged the order of the paragraphs.

policies/coding-style.md
DDvO: I thought that also, e.g., `camelCase` should not be used unless such a name is inherited, e.g., from a standard?

policies/coding-style.md
DDvO: I was told earlier that also names like `e`, `n`, and the like should be avoided,
i.e., that single-letter variable names should be avoided except for loop counters.
What do others think?

DDvO: Apart from being rather ambiguous and not descriptive, 
one-letter names have the drawback that searching for their occurrences (using a normal editor or browser) is a pain.

t8m: local variable names like `e` and `n` are in my opinon very much fine if they are for example about the exponent e and modulus n in RSA.

> I was told earlier that also names like `e`, `n`, and the like should be avoided,

When that was?

policies/coding-style.md
t8m: I'd remove it altogether.

t8m: Hmm, actually the original paragraph was fine and I could understand as supporting `tmp` for a variable name if there is no ambiguity, i.e. only one temporary variable in the function.

t8m: I would even say the original paragraph was just fine and if you just moved it, it would be OK. Because the sentence about `loop_counter` goes well with the previous sentence in the original wording, not so much in your edited version.

policies/coding-style.md
mattcaswell: There are definitely instances in the code where it is confusing, e.g.

https://github.com/openssl/openssl/blob/802cacf34f2db9111becb4f0d3aa00460df13a19/ssl/t1_enc.c#L292-L314

DDvO: >> Apart from being rather ambiguous and not descriptive, 
one-letter names have the drawback that searching for their occurrences (using a normal editor or browser) is a pain.

> local variable names like `e` and `n` are in my opinion very much fine if they are for example about the exponent e and modulus n in RSA.

I agree that in such cases the names are telling.
Still be problem that it's hard to search occurrences stays.

> > I was told earlier that also names like `e`, `n`, and the like should be avoided,
> 
> When that was?

A couple of years back, when I contributed the CMP implementation.
I could try and did those reviewer comments out if this is helpful.

policies/coding-style.md
DDvO: Note that previously there were two paragraphs mentioning the name `tmp`:
```
Do not use long
names like _ThisVariableIsATemporaryCounter_. Use a name like _tmp_, which
is much easier to write, and not more difficult to understand.
```
I removed the this one because it was redundant,
and this is the one you found more easy to understand, and I agree on this.

The second one is this:
```
Local variable names should be short, and to the point. If you have
some random integer loop counter, it should probably be called _i_.
Calling it _loop_counter_ is non-productive, if there is no chance of it
being mis-understood. Similarly, _tmp_ can be just about any type of
variable that is used to hold a temporary value
```
and this is the one I just moved around.

Unfortunately, its last sentence is hard to understand.
How about replacing it by, e.g.:
```
Use brief name name like `tmp` and `name` as long as they are non-ambiguous in the given context.
```

t8m: > How about replacing it by, e.g.:
> 
> ```
> Use brief name name like `tmp` and `name` as long as they are non-ambiguous in the given context.
> ```

That would be fine.

policies/coding-style.md
t8m: > Still be problem that it's hard to search for occurrences stays.

But if it is a local variable, then you should not need to search for it in the whole source code context.

t8m: > There are definitely instances in the code where it is confusing, e.g.
> 
> https://github.com/openssl/openssl/blob/802cacf34f2db9111becb4f0d3aa00460df13a19/ssl/t1_enc.c#L292-L314

That is NOT a short simple function, is it? And those single letter variable instances and even the two-letter ones are clearly not unambiguous.

DDvO: Yes, but even within a (not pretty short) function searching for names like `e` and `x` can be quite a pain.

t8m: In that function I'd still probably keep using `s` as the `SSL *` parameter. Otherwise I am not sure I'd keep any other variable as single letter one an probably also rename a few others.

policies/coding-style.md
DDvO: Done

policies/coding-style.md
DDvO: I've replaced the sentence as suggested and also used the given examples.

policies/coding-style.md
t8m: So I would be OK with `Apart from that, avoid single-letter names if they are ambiguous or when they can be visually confusing...`

policies/coding-style.md
t8m: You did not remove the original sentence.

policies/coding-style.md
DDvO: How about this:
```
Avoid single-letter names when they can be visually confusing, such as `I` and `O`.
Avoid other single-letter names unless they are telling in the given context.
For instance, `m` for modulus and `s` for `SSL` pointers are fine.
```

t8m: Good.

DDvO: Done.

t8m (approved):

DDvO:
@levitte are your above points sufficiently addressed, so can you approve?

policies/coding-style.md
```diff
@@ -234,27 +234,51 @@ changing their context lines.
...
 
-Except when otherwise required, avoid mixed-case names.
+Local variable names should be short, and to the point. If you have
+some random integer loop counter, it should probably be called `i` or `j`.
 
-Do not encode the type into a name (so-called Hungarian notation).
+Avoid single-letter names when they can be visually confusing,
+such as `I` and `O`.
+Avoid other single-letter names unless they are telling in the given context.
+For instance, `m` for modulus and `s` for SSL pointers are fine.
```
slontis, line 245: I dont agree that 's' is a good name

policies/coding-style.md
```diff
@@ -234,27 +234,51 @@ changing their context lines.
...
+
+Calling a variable `loop_counter` is non-productive,
+if there is no chance of it being mis-understood.
```
slontis, line 248: This is probably redundant as you already say use i of j above

policies/coding-style.md
```diff
@@ -234,27 +234,51 @@ changing their context lines.
...
+or it is duplicated/up-ref'ed such that an additional `free()` will be needed.
 
-If you are afraid that someone might mix up your local variable names,
-perhaps the function is too long; see Chapter 6.
+Use lowercase prefix like `ossl_` for internal symbols
+unless they are `static` (i.e., local to the source file).
+
+Use uppercase prefix like `EVP_` or `OSSL_CMP_` for public (API) symbols.
+
+Do not encode the type into a name (so-called Hungarian notation).
```
slontis, line 273: A simple example here might be good i.e.  int iAge;

policies/coding-style.md
mattcaswell: 's' is used very extensively throughout libssl and its meaning is clear there.

t8m:
@DDvO Would you like to address (at least some) @slontis comments?

policies/coding-style.md
DDvO: I agree - so removed this sentence since it is in some sense also covered by the next one:

> Use simple variable names like `tmp` and `name` as long as they are non-ambiguous in the given context.

policies/coding-style.md
DDvO: Added this example.

DDvO:
> @DDvO Would you like to address (at least some) @slontis comments?

Sure - I've just been too busy to do earlier, but done now.

t8m (approved):

mattcaswell (approved):

t8m:
Starting the vote for coding-style.md: add various rules on names at commit [5a3b1ac](https://github.com/openssl/technical-policies/pull/48/commits/5a3b1ac9bcae6f5887df51fe0a3aa252c6ade22c)

t8m:
Vote: [+1]

mattcaswell:
Vote: [+1]

romen:
Vote: [+0]

beldmit:
Vote: [+1]

slontis:
Vote: [+1]

t-j-h:
Vote: [+1]

paulidale:
Vote: [+1]

kroeckx:
Voting +1

levitte:
Vote: [+1]

t8m:
@mspncp ping for vote

mspncp:
Vote: [+1]

t8m:
Closing the vote.

```
Topic: coding-style.md: add various rules on names at commit 5a3b1ac
Proposed by: Tomas
Issue link: https://github.com/openssl/technical-policies/pull/48
Public: yes
Opened: 2022-05-25
Closed: 2022-06-02
Accepted: yes   (for: 9, against: 0, abstained: 1, not voted: 0)

   Dmitry     [+1]
   Matt       [+1]
   Pauli      [+1]
   Tim        [+1]
   Richard    [+1]
   Shane      [+1]
   Tomas      [+1]
   Kurt       [+1]
   Matthias   [+1]
   Nicola     [+0]
```

Issue #33, opened 2022-03-14
Switch to C11 instead of C89/C90


kroeckx:
The main reason we were stuck on C89/C90 is that Microsoft Visual C didn't support anything more recent. Recent versions do support C11: https://docs.microsoft.com/en-us/cpp/overview/install-c17-support?view=msvc-170. Note that they don't support C99. I suggest we consider changing to C11.

t-j-h:
No - the main reason we stick to earlier versions is so that OpenSSL can go across the widest range of platforms and that is not just on Windows. People work on a lot of platforms which are much older and on a pile of platforms where the compiler is much older too. Moving forward would cause more than a few challenges and for little meaningful benefit.

The OpenSSL community isn't in general sitting on the latest compilers on the latest platforms.

h-vetinari:
C11 is the default since GCC 5.x, which can hardly be called the latest compiler (and covers a very wide range of platforms). The linux kernel is also [moving](https://lore.kernel.org/lkml/20220308215615.14183-4-arnd@kernel.org/) to C11 in the upcoming merge window.

Not saying that this is directly applicable to OpenSSL, but the discussion itself makes sense IMO. What supported platforms would be affected and how? From https://www.openssl.org/policies/platformpolicy.html, it looks like probably NonStop (CC @rsbeckerca) and OpenVMS (documented as having some C99 [support](https://docs.vmssoftware.com/vsi-c-language-reference-manual/#Preface), though they seem to be [moving](https://vmssoftware.com/docs/2021-04-27-x86-Compiler-Update.pdf) to an LLVM backend, at least on x86).

This leads to a (potentially much) larger point why OpenSSL would willingly hold itself back to perform free(?) support for commercial entities like HP and VMS Software, but I acknowledge that this is not likely to change. Still, transparency on such constraints (and effective assumption of burden by OpenSSL) are IMO desirable, hence why I think this discussion is worth having.

rsbeckerca:
Moving to c11 will exclude fixes to some legacy platforms still in use and maintained. I am happy to accept moving to c99 as a baseline, but not c11 for a few more years. We still have to support existing users. I do not see what the advantage is to moving to c11 - the functionality in the language is just not that broadly important in a pure C application. I understand a C++ motivation, but not in this situation.

paulidale:
I'm honestly not see much that c11 gives that we'd use: the alignment specifications and maybe anonymous structs and unions.  The third would be atomics but we've got them if compiler supported.

[...]

PR #50, opened 2022-05-17
Switch to C99 without the features that became optional in C11

kroeckx:


policies/coding-style.md
```diff
@@ -586,10 +586,8 @@ it portable enough to meet our multi-platform support goals.
 
 ## Chapter 14:  Portability
 
-To maximise portability the version of C defined in ISO/IEC 9899:1990
-should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
-(also known as C99) is not supported on some platforms that OpenSSL is
-used on and therefore should be avoided.
+The version of C defined in ISO/IEC 9899:1999 (C99) should be used.
+Features that became optional in ISO/IEC 9899:2011 (C11) should not be used.
```
t8m, line 590: Should this change apply to 3.1 and newer releases only?

levitte:
How do we handle old compilers, on platforms where there isn't any choice?  Will it be sufficient to add compatibility functions and macros, or are we thinking of supporting syntactic things that just aren't possible on those (for example, field specifiers when setting the value of a union)?

policies/coding-style.md
kroeckx: That was at least the intention. How do we handle such changes in other policies? just add some text saying it only applies to 3.1 and later?

kroeckx:
We already have compatibility for such platforms. We have plenty of code that's there just because we want to use C99 and C11 features that are not available on all platforms. But as far as I know, we don't have a platform in our platform lists that doesn't support C99, which includes things like Sun C and djgpp. This change in policy will allow us to remove such compatibility.

t-j-h:
I don't believe that there are any C99 features that are sufficiently important to move forward to that language version.
OpenSSL is used on a large range of platforms that go well beyond the official list. We would also have to continue with C89 level support for all current releases.

hlandau:
What are the features in C99 we actually want?

Taking a list of features from Wikipedia:
>inline functions

We already have our own macro for this.

>intermingled declarations and code

Would be nice, but major stylistic change, so probably undesirable.

>several new [data types](https://en.wikipedia.org/wiki/Data_type), including long long int, optional extended integer types, an explicit [boolean data type](https://en.wikipedia.org/wiki/Boolean_datatype), and a complex type to represent [complex numbers](https://en.wikipedia.org/wiki/Complex_number):

Bool would be nice and easily polyfilled on platforms which don't have it. Complex probably has poor support and I can't see us needing it.  long long we already use.

>[flexible array members](https://en.wikipedia.org/wiki/Flexible_array_member)

This doesn't really buy us anything and probably has poor support.

>support for one-line [comments](https://en.wikipedia.org/wiki/Comment_(programming)) beginning with //, as in [BCPL](https://en.wikipedia.org/wiki/BCPL), [C++](https://en.wikipedia.org/wiki/C%2B%2B) and [Java](https://en.wikipedia.org/wiki/Java_(programming_language))

Would be nice, but major stylistic change, so probably undesirable.

>new library functions, such as snprintf

We use our own implementation of printf in BIO, so this doesn't really matter to us.

>designated initializers (for example, initializing a structure by field names: struct point p = { .x = 1, .y = 2 };)[[5]](https://en.wikipedia.org/wiki/C99#cite_note-5)
>compound literals (for instance, it is possible to construct structures in function calls: function((struct x) {1, 2}))[[6]](https://en.wikipedia.org/wiki/C99#cite_note-6)

Would be nice but not a game-changer. I think we would need to survey support on our target platforms (VMS, NonStop, ...)

>support for [variadic macros](https://en.wikipedia.org/wiki/Variadic_macro) (macros with a variable number of arguments)

Probably the most useful thing in C99 for us. This only affects the preprocessor.

>[restrict](https://en.wikipedia.org/wiki/Restrict)

Desirable. We can use a macro for this so we can compile it out on platforms that don't support it. We can adopt this whether or not we adopt C99 in general.

I think aside from creating an `ossl_restrict` macro the only things useful to us are `_Bool` and variadic macros. We could also  create an `ossl_static_assert` macro for platforms which support C11 static assert. Of these the only things which can't be polyfilled/no-opped on platforms which don't support them are variadic macros. So the question is whether all platforms can support them.

beldmit:
Personally I like boolean, declaring cycle variables inside the cycle itself and //-style comments. `long long` looks like a bad idea to me, I prefer explicitly sized types.

policies/coding-style.md
t8m: > just add some text saying it only applies to 3.1 and later?

Yes, I think so.

t8m:
> `long long` looks like a bad idea to me, I prefer explicitly sized types.

We already need and use it on 32 bit platforms for 64 bit numbers. (And yeah mostly as a typedef to int64_t)

levitte:
> > designated initializers (for example, initializing a structure by field names: struct point p = { .x = 1, .y = 2 };)[[5]](https://en.wikipedia.org/wiki/C99#cite_note-5)
> > compound literals (for instance, it is possible to construct structures in function calls: function((struct x) {1, 2}))[[6]](https://en.wikipedia.org/wiki/C99#cite_note-6)
> 
> Would be nice but not a game-changer. I think we would need to survey support on our target platforms (VMS, NonStop, ...)
> 
> > support for [variadic macros](https://en.wikipedia.org/wiki/Variadic_macro) (macros with a variable number of arguments)
> 
> Probably the most useful thing in C99 for us. This only affects the preprocessor.

These two would be useful, but seriously precludes older compilers / pre-processors, as these are syntax features that can't be replicated with smart functions or macros...
That being said, *current* VMS (v8.4) with a *current* C compiler (v7.4) [seems to fully support C99](https://docs.vmssoftware.com/vsi-c-language-reference-manual/)

kroeckx:
My understanding of the current situation is that we currently use those unconditionally:
- long long int
- flexible array members (not sure if we still do)
- snprintf

We use those conditionally:
- inttypes (like uint64_t)
- inline functions

Things that look useful that we currently don't use:
- intermingled declarations and code
- booleans
- designated initializers
- variadic macros
- restrict
- Unicode

Things that some people might want to use:
- one-line comments using //.  The coding style doesn't allow it.

hlandau:
I doubt Unicode is of use to us.

h-vetinari:
> > several new [data types](https://en.wikipedia.org/wiki/Data_type), including long long int, optional extended integer types, an explicit [boolean data type](https://en.wikipedia.org/wiki/Boolean_datatype), and a complex type to represent [complex numbers](https://en.wikipedia.org/wiki/Complex_number):
> 
> Bool would be nice and easily polyfilled on platforms which don't have it. Complex probably has poor support and I can't see us needing it. long long we already use.

There's an issue for that (and some preliminary discussion on C99) already in https://github.com/openssl/openssl/issues/17494.

Also, nonstop would be fine with C99 based on [this](https://github.com/openssl/technical-policies/issues/33#issuecomment-1069580781) comment.

Sidenote: this PR should probably have "Closes #33" in the OP.

kroeckx:
Unicode support is something that can be useful in the apps, but to properly use it we probably also need something like iconv().

kroeckx:
An alternative could be that we list which C99 features that we require.

t8m:
> An alternative could be that we list which C99 features that we require.

**OTC: We think that this is the only way forward. Approving C99 as a whole does not work for supporting older platforms that we still want to support.**

t8m:
@kroeckx ping

t8m:
**OTC: vote has not been called. Closing.**

PR #51, opened 2022-05-18
Deprecate long and add notes on integer types

hlandau:
Fixes https://github.com/openssl/openssl/issues/18338

### DIFF
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -702,10 +702,13 @@ or have important caveats:
   on some platforms.
 
 Existing public APIs using deprecated types must be maintained, but new APIs and
-new code should avoid use of them.
+new code should avoid use of them. Deprecation does not mean that existing code
+need be converted when it is modified, and other code may still use deprecated
+types where necessary to interoperate with that code. However, deprecated types
+should not be used in completely new code.
 
-The above advice does not restrict typedef or macro definitions expressed in
-terms of deprecated integer types made for portability purposes.
+The above advice also does not restrict typedef or macro definitions expressed
+in terms of deprecated integer types made for portability purposes.
 
 ## Chapter 15: Expressions
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -692,7 +692,7 @@ or have important caveats:
 - Use `signed char` only where the range of values required is a subset of
   the range `[-127, 127]`, as this range is guaranteed by the C standard.
 
-  However, consider using `int8_t` instead.
+  As with `char`, consider using `int8_t` instead.
 
 - Use `unsigned char` for byte buffers. `unsigned char` may also be
   used for integers, but consider using `uint8_t` instead.
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -677,7 +677,7 @@ or have important caveats:
 
   The C standard only guarantees that `int` can represent the range
   `[-32767, 32767]`, but it is an assumption of the OpenSSL codebase that `int`
-  is 32 bits.
+  is at least 32 bits in size.
 
 - Use `char` for integers only where the range of values required is a subset of
   the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -683,6 +683,9 @@ or have important caveats:
 
   When using `unsigned int`, write `unsigned int` instead of just `unsigned`.
 
+- `long long` and `unsigned long long` are deprecated; use `int64_t` or
+  `uint64_t` instead.
+
 - Use `char` for integers only where the range of values required is a subset of
   the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
   is unsigned by default on some platforms.)
@@ -699,6 +702,9 @@ or have important caveats:
 
 - Use `char` for strings, but bear in mind that `char` is unsigned by default
   on some platforms.
+
+The above advice does not restrict typedef or macro definitions expressed in
+terms of deprecated integer types made for portability purposes.
 
 ## Chapter 15: Expressions
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -673,10 +673,11 @@ or have important caveats:
 - Use of `short` or `unsigned short` is deprecated. In the worst case, `short` has
   the same size as `int`.
 
-- Use `int` or `unsigned int` only where the range of values required is a subset
-  of the range `[-32767, 32767]`, as this is guaranteed by the C standard.
+- `int` and `unsigned int` can be used.
 
-  Consider using `int32_t` or `uint32_t` instead.
+  The C standard only guarantees that `int` can represent the range
+  `[-32767, 32767]`, but it is an assumption of the OpenSSL codebase that `int`
+  is 32 bits.
 
 - Use `char` for integers only where the range of values required is a subset of
   the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
@@ -689,7 +690,7 @@ or have important caveats:
 
   However, consider using `int8_t` instead.
 
-- Use `unsigned char` or `uint8_t` for byte buffers. `unsigned char` may also be
+- Use `unsigned char` for byte buffers. `unsigned char` may also be
   used for integers, but consider using `uint8_t` instead.
 
 - Use `char` for strings, but bear in mind that `char` is unsigned by default
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -670,8 +670,10 @@ or have important caveats:
   Use `size_t` to represent array lengths and indices. If a signed value is
   needed, use `ossl_ssize_t`.
 
-- Use of `short` or `unsigned short` is deprecated. In the worst case, `short` has
-  the same size as `int`.
+- `short` or `unsigned short` should not be used in public APIs or to denote
+  a width. These types should be used only where the range of values required is
+  a subset of the range `[-32767, 32767]`, as this range is guaranteed by the C
+  standard.
 
 - `int` and `unsigned int` can be used.
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -642,10 +642,58 @@ it portable enough to meet our multi-platform support goals.
 
 ## Chapter 14: Portability
 
+### Language
+
 To maximise portability the version of C defined in ISO/IEC 9899:1990
 should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
 (also known as C99) is not supported on some platforms that OpenSSL is
 used on and therefore should be avoided.
+
+### Integer types
+
+The types `int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`,
+`int64_t` and `uint64_t` are recommended for use. These were only standardised
+in C99 but we define them ourselves if they are not available on a target
+platform.
+
+`size_t` is available. `ossl_ssize_t` should be used instead of `ssize_t` for
+portability.
+
+Most of the built-in integer types in the C language should be avoided
+or have important caveats:
+
+- Use of `long` and `unsigned long` is deprecated, as these types cannot be
+  relied upon to be larger than `int` on all platforms. Existing public APIs
+  which already use these types must be maintained, but new APIs should avoid
+  usage of them. Internal code should avoid using these types.
+
+  Use `size_t` to represent array lengths. If a signed value is needed, use
+  `ossl_ssize_t`.
+
+- Use of `short` or `unsigned short` is deprecated. In the worst case, `short` has
+  the same size as `int`.
+
+- Use `int` or `unsigned int` only where the range of values required is a subset
+  of the range `[-32767, 32767]`, as this is guaranteed by the C standard.
+
+  Consider using `int32_t` or `uint32_t` instead.
+
+- Use `char` for integers only where the range of values required is a subset of
+  the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
+  is unsigned by default on some platforms.)
+
+  However, consider using `int8_t` instead.
+
+- Use `signed char` only where the range of values required is a subset of
+  the range `[-127, 127]`, as this range is guaranteed by the C standard.
+
+  However, consider using `int8_t` instead.
+
+- Use `unsigned char` or `uint8_t` for byte buffers. `unsigned char` may also be
+  used for integers, but consider using `uint8_t` instead.
+
+- Use `char` for strings, but bear in mind that `char` is unsigned by default
+  on some platforms.
 
 ## Chapter 15: Expressions
 
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -652,9 +652,9 @@ used on and therefore should be avoided.
 ### Integer types
 
 The types `int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`,
-`int64_t` and `uint64_t` are recommended for use. These were only standardised
-in C99 but we define them ourselves if they are not available on a target
-platform.
+`int64_t` and `uint64_t` are recommended for use when the exact width is
+important. These were only standardised in C99 but we define them ourselves if
+they are not available on a target platform.
 
 `size_t` is available. `ossl_ssize_t` should be used instead of `ssize_t` for
 portability.
@@ -678,6 +678,8 @@ or have important caveats:
   The C standard only guarantees that `int` can represent the range
   `[-32767, 32767]`, but it is an assumption of the OpenSSL codebase that `int`
   is at least 32 bits in size.
+
+  When using `unsigned int`, write `unsigned int` instead of just `unsigned`.
 
 - Use `char` for integers only where the range of values required is a subset of
   the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -663,9 +663,7 @@ Most of the built-in integer types in the C language should be avoided
 or have important caveats:
 
 - Use of `long` and `unsigned long` is deprecated, as these types cannot be
-  relied upon to be larger than `int` on all platforms. Existing public APIs
-  which already use these types must be maintained, but new APIs should avoid
-  usage of them. Internal code should avoid using these types.
+  relied upon to be larger than `int` on all platforms.
 
   Use `size_t` to represent array lengths and indices. If a signed value is
   needed, use `ossl_ssize_t`.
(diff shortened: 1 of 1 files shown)


policies/coding-style.md
```diff
@@ -586,11 +586,59 @@ it portable enough to meet our multi-platform support goals.
...
+Most of the built-in integer types in the C language should be avoided
+or have important caveats:
+
+- Use of `long` and `unsigned long` is deprecated, as these types cannot be
+  relied upon to be larger than `int` on all platforms. Existing public APIs
+  which already use these types must be maintained, but new APIs should avoid
+  usage of them. Internal code should avoid using these types.
+
+  Use `size_t` to represent array lengths. If a signed value is needed, use
+  `ossl_ssize_t`.
```
levitte, line 615: We also use `size_t` for array indexing

policies/coding-style.md
```diff
@@ -586,11 +586,59 @@ it portable enough to meet our multi-platform support goals.
...
+
+- Use of `short` or `unsigned short` is deprecated. In the worst case, `short` has
+  the same size as `int`.
+
+- Use `int` or `unsigned int` only where the range of values required is a subset
+  of the range `[-32767, 32767]`, as this is guaranteed by the C standard.
```
t8m, line 621: That's wrong. We depend on 32 bit int/unsigned int and IMO we can continue to do so. I would seriously dislike to see proliferation of int32_t everywhere.

policies/coding-style.md
```diff
@@ -586,11 +586,59 @@ it portable enough to meet our multi-platform support goals.
...
+  is unsigned by default on some platforms.)
+
+  However, consider using `int8_t` instead.
+
+- Use `signed char` only where the range of values required is a subset of
+  the range `[-127, 127]`, as this range is guaranteed by the C standard.
+
+  However, consider using `int8_t` instead.
+
+- Use `unsigned char` or `uint8_t` for byte buffers. `unsigned char` may also be
```
t8m, line 636: As a style issue, I'd keep using unsigned char for byte buffers and not start using uint8_t. That should be used only where it is used for unsigned integers of that size.

hlandau:
Updated.

policies/coding-style.md
levitte: This is not wrong.  16 bits _is_ the minimum size for ints guaranteed by the C standard.

Now, *we* have chose to require a minimum 32-bit size for int, thus basically discarding any 16-bit system (so long, DOS, you won't be missed :wink:)...  but that's not what this sentence says.

policies/coding-style.md
```diff
@@ -617,10 +617,11 @@ or have important caveats:
...
   the same size as `int`.
 
-- Use `int` or `unsigned int` only where the range of values required is a subset
-  of the range `[-32767, 32767]`, as this is guaranteed by the C standard.
+- `int` and `unsigned int` can be used.
 
-  Consider using `int32_t` or `uint32_t` instead.
+  The C standard only guarantees that `int` can represent the range
+  `[-32767, 32767]`, but it is an assumption of the OpenSSL codebase that `int`
+  is 32 bits.
```
t8m, line 624: I'd say `at least 32 bits long`.  `int32_t` or `uint32_t` should be used only in places where the value must be exactly 32 bit long.

hlandau:
Updated.

hlandau:
Updated.

policies/coding-style.md
```diff
@@ -586,11 +586,62 @@ it portable enough to meet our multi-platform support goals.
...
+
+- Use of `long` and `unsigned long` is deprecated, as these types cannot be
+  relied upon to be larger than `int` on all platforms. Existing public APIs
+  which already use these types must be maintained, but new APIs should avoid
+  usage of them. Internal code should avoid using these types.
+
+  Use `size_t` to represent array lengths and indices. If a signed value is
+  needed, use `ossl_ssize_t`.
+
+- Use of `short` or `unsigned short` is deprecated. In the worst case, `short` has
```
paulidale, line 617: I'm not convinced.  Sure, `short int` might be the same size as `int` on some platforms but it is narrower on many.  Using `short int` is an indication that the full range of `int` isn't required, being wider is harmless.

I do agree they shouldn't be used in public APIs or to denote a width.

policies/coding-style.md
```diff
@@ -586,11 +586,64 @@ it portable enough to meet our multi-platform support goals.
...
+- Use `char` for integers only where the range of values required is a subset of
+  the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
+  is unsigned by default on some platforms.)
+
+  However, consider using `int8_t` instead.
+
+- Use `signed char` only where the range of values required is a subset of
+  the range `[-127, 127]`, as this range is guaranteed by the C standard.
+
+  However, consider using `int8_t` instead.
```
mattcaswell, line 639: This is a repeat of line 634. It looks odd.

hlandau: Slight tweak

policies/coding-style.md
```diff
@@ -586,11 +586,64 @@ it portable enough to meet our multi-platform support goals.
...
+important. These were only standardised in C99 but we define them ourselves if
+they are not available on a target platform.
+
+`size_t` is available. `ossl_ssize_t` should be used instead of `ssize_t` for
+portability.
+
+Most of the built-in integer types in the C language should be avoided
+or have important caveats:
+
+- Use of `long` and `unsigned long` is deprecated, as these types cannot be
```
paulidale, line 609: The _exact_ same arguments for not using `long` also apply to `long long`.  It's is only guaranteed to be as wide as an `int`.

t8m: This is debatable. IMO we depend on long long being 64 bit and that is held on all platforms we currently support. Of course we could replace long long with int/uint64_t except with BIO_snprintf it is a little bit awkward to use `%lld` with cast to `(int64_t)`. So maybe we should deprecate long long and unsigned long long except for use with printf.

paulidale: Existing uses have to remain of course, but I stand by my statement that the **exact** same arguments against `long` also apply to `long long`.  If we're deprecating the former, we should also deprecate the latter.  Regardless of the problems it might cause.

Yes, `BIO_printf` et al are problematic, yet unmentioned here.  They have zero scope for supporting the fixed sized integers.  Their supporting `size_t` was (mildly) annoying when the `z` modifier was added.

levitte: We should probably use the PRI macros a bit more consistently

https://en.cppreference.com/w/c/types/integer

paulidale: This would remove my objection to deprecating `short int`.

They aren't in C89 and we'd need to implement wrappers if we don't accept C99.

t8m: I actually think we should start small and just deprecate long and unsigned long, for others we just state the current limits we depend on in the code base.

levitte: Nah, I'd rather not do this policy change incrementally.  I think it's safe enough if we have the (stated?) expectation that this applies primarily to new code, leaving the option of modifying old code exactly that, an option.

hlandau: I just checked and long long has to be at least 64 bits in both C99 and C11. But there's no reason to use it directly, so let's deprecate it.

The correct way to use fixed width integer types with printf is with the macros defined in inttypes.h:

```c
printf("Value: " PRIu32 "\n", (uint32_t)42);
```

This is admittedly painful. Though I think we may have to go here anyway as I've encounted a great volume of warnings when targeting mingw about not supporting `%zu`, which we use extensively.

hlandau:
Updated.

policies/coding-style.md
levitte: > This is admittedly painful. Though I think we may have to go here anyway as I've encounted a great volume of warnings when targeting mingw about not supporting `%zu`, which we use extensively.

It's only very recently that MSVC *stopped* giving those same warnings...

paulidale: We're using C89 currently....

levitte: > We're using C89 currently....

... which is clear with our already liberal use of [u]int*nn*_t :wink:

hlandau: long long doesn't exist in C89.

hlandau:
Updated.

policies/coding-style.md
```diff
@@ -646,10 +646,13 @@ or have important caveats:
   on some platforms.
 
 Existing public APIs using deprecated types must be maintained, but new APIs and
-new code should avoid use of them.
+new code should avoid use of them. Deprecation does not mean that existing code
+need be converted when it is modified, and other code may still use deprecated
```
t8m, line 650: Typo: needs

hlandau: Either wording is OK with me, but this is grammatical.

policies/coding-style.md
```diff
@@ -586,11 +586,74 @@ it portable enough to meet our multi-platform support goals.
...
+- Use `signed char` only where the range of values required is a subset of
+  the range `[-127, 127]`, as this range is guaranteed by the C standard.
+
+  As with `char`, consider using `int8_t` instead.
+
+- Use `unsigned char` for byte buffers. `unsigned char` may also be
+  used for integers, but consider using `uint8_t` instead.
+
+- Use `char` for strings, but bear in mind that `char` is unsigned by default
+  on some platforms.
```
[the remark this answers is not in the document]
kroeckx: Or should we define that all strings in public APIs are in UTF8?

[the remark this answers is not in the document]
kroeckx: Note that C11 allows `u8"string"` and that represents an UTF8 string stored as unsigned char. We should probably use unsigned char strings and deprecate char.

I  understand that a future C standard (C2X) will introduce a char8_t that is a distinct type that works with u8.

hlandau:
Rebased.

mattcaswell:
@hlandau - there seems to be some unresolved comments from @kroeckx. Can you respond before we take this to a vote?

policies/coding-style.md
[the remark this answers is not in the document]
hlandau: We don't use `wchar_t` anywhere, I don't think it bears talking about.

String encoding is an attribute of the public API, not coding style, so I don't think it belongs here.

I don't think using unsigned char strings is realistic or desirable. Some platforms have signed char, other platforms have unsigned char, in either case those are what OS APIs expect to receive. It would be an overbearing change to the codebase, and should be discussed in a separate proposal.

policies/coding-style.md
```diff
@@ -586,11 +586,74 @@ it portable enough to meet our multi-platform support goals.
...
+  When using `unsigned int`, write `unsigned int` instead of just `unsigned`.
+
+- `long long` and `unsigned long long` are deprecated; use `int64_t` or
+  `uint64_t` instead.
+
+- Use `char` for integers only where the range of values required is a subset of
+  the range `[0, 127]`, as this range is guaranteed by the C standard. (`char`
+  is unsigned by default on some platforms.)
+
+  However, consider using `int8_t` instead.
```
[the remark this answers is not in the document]
hlandau: I don't think it's a good idea to start dealing with aliasing concerns in this PR. Perhaps make a separate proposal?

hlandau:
@mattcaswell I've responded, I don't think changes are needed.

policies/coding-style.md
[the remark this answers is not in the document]
kroeckx: This policy is not about how we talk to the OS. This is how applications talk to the library. We currently do not say anything about encoding of characters. As a result, you can't write applications that properly display the strings. Our own apps just write something to the console and hope the user can read it. One way of dealing with that is is saying that it needs to be UTF8 encoded. An other way of dealing with it is using something like wchar_t.

You can argue that string encoding is not a coding style, but you're now placing it in our coding style.

We should not be directly writing strings to the OS, so I don't see a problem with it not being a char.

policies/coding-style.md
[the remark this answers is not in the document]
kroeckx: All I'm saying is that we should not be suggesting that people use char, they should just use int8_t or uint8_t if that is what they need.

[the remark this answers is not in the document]
hlandau: This does not make sense, OS APIs accept `char *`. We cannot use `uint8_t` or `int8_t` for text as this would be incompatible with the OS.

policies/coding-style.md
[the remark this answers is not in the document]
hlandau: This PR doesn't say anything about string encoding.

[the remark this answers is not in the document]
kroeckx: And that is the problem. This does not improve anything. All new APIs will still have the same problem that it has some unknown encoding.

policies/coding-style.md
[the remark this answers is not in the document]
kroeckx: This is about small numbers, not text.

policies/coding-style.md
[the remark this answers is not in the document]
hlandau: If we want to have a technical policy on string encoding that's fine, but that is completely orthogonal to integer types and belongs in a separate PR.

I see limited scope for a policy on string encoding since we interact with the OS using functions like `fopen` etc. and thus essentially inherit the OS's string encoding. We could have a policy which basically says nothing but "our policy is the OS's policy" but material improvements seem unavailable. "Everything must be UTF-8" is a non-starter.

policies/coding-style.md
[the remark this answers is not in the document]
hlandau: Personally I would prefer to see use of `int8_t` and `uint8_t` rather than char. However, this is a somewhat larger and more jarring change to the codebase. For example we already tend to use `unsigned char *` to refer to binary buffers, and while I would support a change to `uint8_t *` in new code, that's a more substantial change than this PR, which I deliberately tried to limit to the most uncontroversial things.

Having said that if there are people who would support a move to `uint8_t *` in new code, do let me know.

hlandau:
Still current, possible candidate for OTC discussion tomorrow.

mattcaswell:
I have an open action to start a vote on this. I can start that shortly if you wish?

t8m:
I do not see a need to discuss this again at OTC meeting. It was already discussed and there were no substantial changes since. So IMO the vote can be started.

hlandau:
Sounds good.

mattcaswell:
Starting the vote for this: deprecate long and add notes on integer types

mattcaswell:
Vote: [+1]

policies/coding-style.md
```diff
@@ -642,11 +642,74 @@ it portable enough to meet our multi-platform support goals.
...
+
 To maximise portability the version of C defined in ISO/IEC 9899:1990
 should be used. This is more commonly referred to as C90. ISO/IEC 9899:1999
 (also known as C99) is not supported on some platforms that OpenSSL is
 used on and therefore should be avoided.
 
+### Integer types
+
+The types `int8_t`, `uint8_t`, `int16_t`, `uint16_t`, `int32_t`, `uint32_t`,
+`int64_t` and `uint64_t` are recommended for use when the exact width is
```
tomato42, line 655: what about `uint128_t`? It would be useful if only for handling result of `uint64_t` multiplication on 64 bit platforms...

beldmit: AFAIR `uint128_t` is hardly even standardized and not supported by all the compilers.

tomato42: gcc and clang support them, gcc has for something like 10 years now (`__uint128_t`)
That being said, MSVC only has the "return the high 64 bits intrinsic", no 128 bit type.

To be clear, I'm not saying that it should be always available, there is no point in it on 32 bit system for one, and there's little point when it can't be used for multiplication...

h-vetinari: > AFAIR `uint128_t` is hardly even standardized and not supported by all the compilers.

It's not standardized because ABI. Relevant [background](https://github.com/openssl/openssl/pull/15830#issuecomment-870474738)

t-j-h:
Vote: [-1]

The recommendation on using int8_t is enough for me to vote against this suggestion.
I also see no reason to avoid using long and using size_t for array indexes we know are *small* makes no sense.

policies/coding-style.md
```diff
@@ -642,11 +642,74 @@ it portable enough to meet our multi-platform support goals.
...
+
+- Use of `long` and `unsigned long` is deprecated, as these types cannot be
+  relied upon to be larger than `int` on all platforms.
+
+  Use `size_t` to represent array lengths and indices. If a signed value is
+  needed, use `ossl_ssize_t`.
+
+- `short` or `unsigned short` should not be used in public APIs or to denote
+  a width. These types should be used only where the range of values required is
+  a subset of the range `[-32767, 32767]`, as this range is guaranteed by the C
```
paulidale, line 673: Hmmm, we make assumptions about running on a 2's complement machine in various places.  This would push us beyond the standard's minimum ranges.

paulidale, line 673: Hmmm, we make assumptions about running on a 2's complement machine in various places.  This would push us beyond the standard's minimum ranges.

t8m:
Vote: [+1]

paulidale:
Vote: [-1]

levitte:
From the responses, it looks like this isn't really ready, and needs another round of fiddling.

Vote: [0]

levitte:
Re size_t for all array indexes, I would say that consistency is a factor. It makes no sense to have different index types depending on the size of the array (what would be the pivot size?)

kroeckx:
I'm voting +1

t8m:
> Re size_t for all array indexes, I would say that consistency is a factor. It makes no sense to have different index types depending on the size of the array (what would be the pivot size?)

Hmm, I actually agree with Tim that size_t should be used only in case we do not know the index will be small. Pivot size? Anything that clearly fits into 31 bits of the guaranteed int size?

t8m:
Anyway I still think this is an improvement for start and can be adjusted further in further PRs.

romen:
Vote: [+0]

beldmit:
I'd like to see some amendments according to the discussion. I strongly like the approach as a whole.
[+0]

mspncp:
Vote: [0]

t8m:
Now it's on @slontis to determine the outcome of the vote.

slontis:
Vote: [-1]
I think it probably goes too far - it needs to be simpler to pass.

mattcaswell:
I have now closed this vote. The final result was:

````
Topic: Deprecate long and add notes on integer types
Proposed by: Matt Caswell
Issue link: https://github.com/openssl/technical-policies/pull/51
Public: yes
Opened: 2022-07-25
Closed: 2022-08-02
Accepted:  no  (for: 3, against: 3, abstained: 4, not voted: 0)

  Dmitry     [+0]
  Matt       [+1]
  Pauli      [-1]
  Tim        [-1]
  Richard    [ 0]
  Shane      [-1]
  Tomas      [+1]
  Kurt       [+1]
  Matthias   [ 0]
  Nicola     [+0]
````

mattcaswell:
It seems to me that there would still be some support within OTC for some variation of this PR, so I will leave this open for now to enable OTC to determine next steps.

policies/coding-style.md
tmshort: Also, `[-32767, 32767]` (really `[-32768, 32767]` does not represent the range of `unsigned short`.

PR #67, opened 2023-04-04
Allow FIPS compliance fixes to be bug fixes

t8m:
on the condition they don't break existing use cases.

### DIFF
--- a/policies/stable-release-updates.md
+++ b/policies/stable-release-updates.md
@@ -46,6 +46,11 @@ existing public specifications that we claim to conform to. Where the
 documentation explicitly contradicts the specifications the documented behavior
 is what matters.
 
+A **bug fix** is also a fix of a current or near future FIPS compliance
+requirement in the FIPS module code if it does not break existing use-cases.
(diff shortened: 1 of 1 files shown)


policies/stable-release-updates.md
```diff
@@ -46,6 +46,11 @@ existing public specifications that we claim to conform to. Where the
 documentation explicitly contradicts the specifications the documented behavior
 is what matters.
 
+A **bug fix** is also a fix of a current or near future FIPS compliance
+requirement in the FIPS module code if it does not break existing use-cases.
```
beldmit, line 50: `if it does not break existing use-cases` looks like too strict here.

t8m: Why? This is what we discussed last week on OTC AFAIK. If we need to break it we can still explicitly give exception.

But I assume in most if not all cases we can simply make it configurable.

paulidale (approved):

t8m:
We can handle this by the normal exception process so I am retracting this.

PR #69, opened 2023-07-05
Add vote for PR 21363: fips complaince on installation

paulidale:
Note: the security policy for the 140-3 validation will mandate the `-pedantic` option as part of the installation instructions.

### DIFF
--- a/votes/20230706-accept-21363.txt
+++ b/votes/20230706-accept-21363.txt
@@ -1,0 +1,20 @@
+Topic: Accept PR #21363 into 3.1 subject to the normal review process
+comment: Making FIPS compliance the default for "make fips_install".  The
+         fipsinstall command still requires -pedantic for compliance.
+Proposed by: Pauli
+Issue link: https://github.com/openssl/technical-policies/pull/69
+Public: yes
+Opened: 2023-07-06
+Closed: 2023-07-07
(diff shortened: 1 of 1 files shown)


paulidale:
Calling a vote to accept openssl/openssl#21363 subject to the usual review process

paulidale:
vote: +1

slontis:
vote: +1

beldmit:
vote: +1

I should say that enforcing EMS. significantly breaks backward compatibility and there will be gazillion of questions "why it doesn't work and how to make it work without breaking FIPS compliance"

mspncp:
vote: +1

mattcaswell:
vote: +1

hlandau:
vote: +1

paulidale:
> I should say that enforcing EMS. significantly breaks backward compatibility and there will be gazillion of questions "why it doesn't work and how to make it work without breaking FIPS compliance"

Yeah, it's not ideal.  I think the questions about a non-compliant installation by default will be more painful.  I don't know if we'd get any but worse would be if someone noticed that we were non-compliant and instead of failing, things worked.

paulidale:
Closing vote which was accepted 6 for, 3 not voted.

t-j-h:
Vote [+0]

t8m:
Vote: [+1]

PR #94, opened 2024-05-20
Add Doxygen comment requirement to coding style

nhorman:
In an effort to better document our code in such a way that is condusive to future developers getting involved with our code base, I thought I would propose adding a requirement to our coding style that structures/macros/functions/etc should have Doxygen style comments added to them, so that we could build alternate docs directly from our source base.

### DIFF
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -484,25 +484,25 @@ code as a guideline:
 
 /**
  * \file doxysample.c
- * This is an brief file description that you may add
+ * This is a brief file description that you may add
  * Subsequent lines contain more detailed information about what you will
- * find defined in this file.  Its not currently required that you add a file
- * description, but its available if you like
+ * find defined in this file.  It is not currently required that you add a file
+ * description, but its available if you like.
  */
 
  /**
   * \def MAX(x, y)
-  * document a macro that returns the maximum of two inputs
+  * document a macro that returns the maximum of two inputs.
   * \param x integer input value
-  * \param y interger input value
+  * \param y integer input value
   * \returns the maximum of x and y
   */
   #define MAX(x, y) (x > y ? x : y)
 
 /**
  * \struct foo_st
- * \brief description of the foo_st struct
- * Optional more detailed description here
+ * \brief description of the foo_st struct.
+ * Optional more detailed description here.
  */
 typedef foo_st {
     int a; /**< Describe the a field here */
@@ -511,8 +511,8 @@ typedef foo_st {
 
 
 /**
- * \brief Describe the function add briefly
- * Add a more detailed description here, like sums two inputs and returns the
+ * \brief Describe the function add briefly.
+ * Add a more detailed description here, like sums two inputs and returns the.
  * results
  * \param a - input integer to add
  * \param b - input integer to add
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -487,7 +487,7 @@ code as a guideline:
  * This is a brief file description that you may add
  * Subsequent lines contain more detailed information about what you will
  * find defined in this file.  It is not currently required that you add a file
- * description, but its available if you like.
+ * description, but it's available if you like.
  */
 
  /**
@@ -512,8 +512,8 @@ typedef foo_st {
 
 /**
  * \brief Describe the function add briefly.
- * Add a more detailed description here, like sums two inputs and returns the.
- * results
+ * Add a more detailed description here, like sums two inputs and returns the
+ * results.
  * \param a - input integer to add
  * \param b - input integer to add
  * \returns the sum of a and b
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -464,6 +464,11 @@ derived types. To this end, use just one data declaration per line (no
 commas for multiple data declarations). This leaves you room for a small
 comment on each item, explaining its use.
 
+In an effort to better translate our source code into documentation that is more
+easily understandable to future developers, please also consider for adding
+Doxygen style comments to any function/data structures/macros/etc that you alter
+or create in the development of patches for OpenSSL.
+
 ## Chapter 9: Macros and Enums
 
 Names of macros defining constants and labels in enums are in uppercase:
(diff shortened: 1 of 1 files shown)


policies/coding-style.md
```diff
@@ -464,6 +464,11 @@ derived types. To this end, use just one data declaration per line (no
 commas for multiple data declarations). This leaves you room for a small
 comment on each item, explaining its use.
 
+In an effort to better translate our source code into documentation that is more
+easily understandable to future developers, please also consider for adding
```
paulidale, line 468: `d/for/`

paulidale (commented):
I think that this is optional should be more strongly mentioned.

Unless, the OTC wants to make it mandatory.  But given the massive number of gaps our APIs would then have....

policies/coding-style.md
```diff
@@ -464,6 +464,11 @@ derived types. To this end, use just one data declaration per line (no
...
+Doxygen style comments to any function/data structures/macros/etc that you alter
+or create in the development of patches for OpenSSL.
```
levitte, line 470: Does this apply to internal / static things too?  I assume this is mostly to be done in header files...

paulidale: I wouldn't expect so.  Doxygen is meant for users not developers.
That doesn't mean we cannot document our internals, however I believe that doc/internal is a better place.

levitte: So the point with what I said is that maybe it should be specified somewhat clearly, something along the lines of "This is expected to happen primarly for ... and primarly in ...".  This is a policy after all.

nhorman: @levitte honestly I think thats up for debate.  @vavroch2010 and I were discussing this as a means for making all of our code a bit more accessible to new developers, so I could see starting to document all our code in this manner as beneficial.  I just thought I would propose it as a means to start debate.  That said, @paulidale is correct.  Typically, Doxygen is used for public apis, though thats in no way a requirement

t8m:
Actually for doc-within-the-sources, isn't it much more useful for the developers of OpenSSL than for the users? Or in other words, why is Doxygen better suited for developers of apps using OpenSSL than the current manual page format?

From this it would make much more sense to me to document internals with Doxygen and keep the current .pod files format for the public API documentation.

If this is really intended to be used for public API documentation, what is the plan for converting the existing .pod files? Does it make sense to have public API documentation in two formats with possibly duplicate information in them?

nhorman:
@t8m my intent was for this to be useful to future developers of openssl, not end users or developers of apps using openssl.  My observation has been that one of the barriers to entry in working on openssl.is that, while the public apis are very well documented, the internals are left in a state in which "the code is the documentation" which is difficult given the complexity of our internals and creates s high bar for a new developer to get started.  This change was meant to propose a method by which we could slowly start more explicitly documenting how our internal functions work to ease that burden.

t8m:
> @t8m my intent was for this to be useful to future developers of openssl, not end users or developers of apps using openssl. My observation has been that one of the barriers to entry in working on openssl.is that, while the public apis are very well documented, the internals are left in a state in which "the code is the documentation" which is difficult given the complexity of our internals and creates s high bar for a new developer to get started. This change was meant to propose a method by which we could slowly start more explicitly documenting how our internal functions work to ease that burden.

OK, so this is for documenting internals. That makes much sense to me. Could you please perhaps adjust the wording to make it more clear?

nhorman:
gladly

kroeckx:
Isn't it already our policy to add/update documentation, including internal? 

The only change would be that if it's written as comment, we now specify a format.

nhorman:
@kroeckx yes, you could make that argument.  The documentation policy is located [here](https://www.openssl.org/policies/technical/documentation-policy.html) separate from the coding style policy, and we seem to follow it fairly well for public symbols (ostensibly because we have a ci job that checks those), but we don't follow it for internal symbols very well.  This is an attempt to:
1) Make the documentation requirement more visible
2) define a format for it to make it more generally useful for future developers getting started in our code base

policies/coding-style.md
```diff
@@ -484,25 +484,25 @@ code as a guideline:
 
 /**
  * \file doxysample.c
- * This is an brief file description that you may add
+ * This is a brief file description that you may add
  * Subsequent lines contain more detailed information about what you will
- * find defined in this file.  Its not currently required that you add a file
- * description, but its available if you like
+ * find defined in this file.  It is not currently required that you add a file
+ * description, but its available if you like.
```
t8m, line 490: `it's`

t8m (approved):

t8m:
**This is ready for vote now.**

t8m:
Voting: 0

levitte:
Vote: 0

mattcaswell (approved):

mattcaswell:
Vote: +1

paulidale:
-1

rsbeckerca:
Just a suggestion on this. My team sets a specific form for doxygen with specific elements so that the API documentation of our code can be derived from doxygen. This allows us to keep the API documentation near the code and in the same file - it seems to allows us to more easily keep the documentation consistent with the code.

t8m:
@kroeckx @romen @t-j-h @slontis @beldmit please vote

t-j-h:
Vote: +1

But I would select the '@' rather than '\\' indicator as it is much more "readable" and common.

beldmit:
Vote: 0

slontis:
Vote: +1
I also prefer @ notation.

t8m:
Still missing votes from @kroeckx and @romen

kroeckx:
Voting +1

t8m:
Vote is now closed, the policy change was accepted.

t8m:
@nhorman please open a new pull request with the [8061a09](https://github.com/openssl/technical-policies/pull/94/commits/8061a096665f8d5864cc08059e9f7e149d6e3ab2) commit. This needs to be separately reviewed as minor edit.

nhorman:
style fix available in https://github.com/openssl/technical-policies/pull/98

PR #107, opened 2024-11-26
make markdown markup consistent

quarckster:


nhorman (approved):
I'm fine with these changes, but the # style differs from the style we use in the openssl repo, do we want to be consistent in all our repositories, or does it not matter?

levitte (approved):

levitte:
Made this a minor edit, 'cause it is

quarckster:
> similar to Neil. Looks good, although keen to learn the motivation.

The motivation is consume the content of these repositories during the website building and to have less hacks in the build script.

t8m (approved):

t8m:
Merged. Thank you.

PR #115, opened 2025-10-15
- update chapter 14 w.r.t. switch from c90 to c99

Sashan:
not sure if it follows the proper update process.

back in May/Jun we agreed to use c99. If I remember correct we never discussed if C99 should be allowed on release branches. I feel we branches 3.5 and older should stick to c90 master and 3.6 can should use c99.

### DIFF
--- a/policies/coding-style.md
+++ b/policies/coding-style.md
@@ -706,9 +706,10 @@ it portable enough to meet our multi-platform support goals.
 ## Chapter 14: Portability
 
 To maximise portability the version of C defined in ISO/IEC 9899:1999
-should be used. This is more commonly referred to as C99. The C99 should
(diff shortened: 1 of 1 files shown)


levitte (approved):

t8m (changes requested):

t8m (approved):

levitte:
At this point, this isn't relevant any more.

Also, it seems like the OpenSSL source is going to get a file `STYLE.md` or something like that (which makes quite a lot of sense), so this policy should probably just be removed, or should refer to that file.

Sashan:
I agree this can be closed. once things in repo will settle down then we can update policy to refer to documentation in repository.

PR #118, opened 2026-04-18
Adds a policy for removal of insecure crypto protocols and primitives.

fwh-dc:
Derived from the discussion in https://openssl-communities.org/d/x2S0uIwN/openssl-mission-vs-legacy-systems

### DIFF
--- a/policies/obsolete-or-insecure-functionality.md
+++ b/policies/obsolete-or-insecure-functionality.md
@@ -1,17 +1,17 @@
-OpenSSL policy for obsoleting cryptographic primitives and protocols
---------------------------------------------------------------------
+OpenSSL Library policy for obsoleting cryptographic primitives and protocols
+----------------------------------------------------------------------------
 
-This formal policy outlines the OpenSSL Project’s framework for managing the
+This formal policy outlines the OpenSSL Library’s framework for managing the
 lifecycle of cryptographic primitives, protocols, and structural features. 
 It ensures a "secure by default" posture while providing a predictable migration
 path for the global ecosystem.
 
-## 1. Policy Objective
-To maintain the integrity of the OpenSSL library by systematically deprecating
+## Policy Objective
+To maintain the integrity of the OpenSSL Library by systematically deprecating
 and removing features that no longer meet modern cryptographic standards or
 industry best practices.
 
-## 2. Triggers for end-of-life
+## Triggers for end-of-life
 Features enter the deprecation lifecycle when they meet one of two criteria:
 
 * **Insecure:** Primitives or features with known practical attacks, or where 
@@ -19,10 +19,10 @@ Features enter the deprecation lifecycle when they meet one of two criteria:
     * Any primitive with an effective security strength below **112 bits** 
       (OpenSSL Security Level 2) is considered insecure for modern use.
     * Protocol features that are exploitable by design.
-* **Obsolete:** Features that remain mathematically sound but have been 
+* **Obsolete:** Features that remain cryptographically sound but have been 
   superseded by more secure or efficient alternatives.
 
-## 3. The end-of-life stages
+## The end-of-life stages
 Removal of cryptographic primitives and protocols happens through several stages
 to conform to OpenSSL API guarantees.
 
@@ -33,22 +33,24 @@ to conform to OpenSSL API guarantees.
   demonstrating a weakness (e.g., Logjam, CRIME) is published.
 * **OpenSSL Action:** 
   * Documentation is updated to reflect industry obsoletion status.
-  * Record a decision that defines when deprecation is allowed.
+  * Record a decision that defines when deprecation is planned.
 
 ### Stage 2: OpenSSL Deprecated (Transition)
-* **Status:** Feature is moved to the **Legacy Provider**.
-* **Trigger:** A new major or minor OpenSSL release.
-* **OpenSSL Action:**:
+* **Status:** Feature needs to be explicitly enabled.
+* **Trigger:** It has been decided to formally deprecate the feature.
+* **OpenSSL Action:**
+  * Support for compile-time disabling the feature (`./config no-<feature>`) is
+    added in a minor release.
+  * Major release: Feature is compile-time disabled by default.
+  * If applicable: Adding a migration path in the ossl-migration-guide.md.
+  * Record a decision that defines when removal is planned.
+  * For cryptographic algorithms:
+     * Major release: Feature moved to the Legacy Provider.
+     * Major (in severe cases: minor) release: Security Levels are adjusted to
+       prohibit the feature at Level 2 or higher.
+  * For other features:
     * Minor release: Compiler attributes (`OSSL_DEPRECATEDIN`) trigger
       warnings for developers.
-    * Minor release: Support for compile-time disabling the feature 
-      (`./config no-<feature>`).
-    * Major release: Feature is compile-time disabled by default
-    * Major release: Feature moved to the Legacy Provider.
-    * Major release: Security Levels are adjusted to prohibit the feature at
-      Level 2 or higher.
-    * Record a decision that defines when removal is allowed.
-    * If applicable: Adding a migration path in the ossl-migration-guide.md.
 
 ### Stage 3: OpenSSL Obsolete (Removal)
 * **Status:** Code is deleted.
@@ -57,18 +59,3 @@ to conform to OpenSSL API guarantees.
 * **OpenSSL Action:** 
   * Removal of source code, headers, and tests.
   * Notice in CHANGES.md.
-
-## 4. Specific Criteria for Protocol Features
-Structural features (extensions, compression, etc.) are evaluated differently
-than mathematical primitives:
-
-1.  **Information Leakage:** Any feature that leaks plaintext length or metadata
-    (e.g., **TLS Compression**) is moved immediately to Stage 2.
-2.  **Forward Secrecy:** Features that rely on static keying material (e.g., 
-    **Static RSA**) are marked Obsolete and moved to Stage 2.
-3.  **Side-Channel Resistance:** If a feature cannot be implemented in
-    "constant-time" on modern hardware, it is prioritized for Stage 3 removal.
-
-> **Emergency Override:** The OpenSSL Foundation Board of Directors reserves the
-> right to bypass Stages 1 and 2 for "Zero-Day" catastrophic breaks, disabling
-> the feature in a minor security patch.
--- a/policies/obsolete-or-insecure-functionality.md
+++ b/policies/obsolete-or-insecure-functionality.md
@@ -8,7 +8,7 @@ path for the global ecosystem.
 
 ## 1. Policy Objective
 To maintain the integrity of the OpenSSL library by systematically deprecating
-and removing components that no longer meet modern cryptographic standards or
+and removing features that no longer meet modern cryptographic standards or
 industry best practices.
 
 ## 2. Triggers for end-of-life
@@ -20,36 +20,40 @@ Features enter the deprecation lifecycle when they meet one of two criteria:
       (OpenSSL Security Level 2) is considered insecure for modern use.
     * Protocol features that are exploitable by design.
 * **Obsolete:** Features that remain mathematically sound but have been 
-  superseded by more secure or efficient alternatives (e.g., TLS 1.2 in the 
-  presence of TLS 1.3), or those that introduce unnecessary "state machine"
-  complexity.
+  superseded by more secure or efficient alternatives.
 
 ## 3. The end-of-life stages
 Removal of cryptographic primitives and protocols happens through several stages
-to conform to OpenSSL api guarantees.
+to conform to OpenSSL API guarantees.
 
 ### Stage 1: Industry Obsolete (Signal)
 * **Status:** Feature is still available and enabled by default.
-* **Trigger:** An industry body (IETF, NIST, etc.) publishes a "Deprecated" or
-  "Historic" status, or a significant research paper (e.g., Logjam, CRIME) is
-  published.
+* **Trigger:** An industry standards body (IETF, NIST, etc.) publishes a
+  "Deprecated" or "Historic" status, or a significant research paper 
+  demonstrating a weakness (e.g., Logjam, CRIME) is published.
 * **OpenSSL Action:** 
   * Documentation is updated to reflect industry obsoletion status.
-  * A Github issue or pull request is created for the actual deprecation.
+  * Record a decision that defines when deprecation is allowed.
 
 ### Stage 2: OpenSSL Deprecated (Transition)
 * **Status:** Feature is moved to the **Legacy Provider**.
-* **Trigger:** A new Major or Minor OpenSSL release.
+* **Trigger:** A new major or minor OpenSSL release.
 * **OpenSSL Action:**:
-    * Feature is disabled by default; must be explicitly loaded via configuration.
-    * Compiler attributes (`OSSL_DEPRECATEDIN`) trigger warnings for developers.
-    * Security Levels are adjusted to prohibit the feature at Level 2 or higher.
-    * An issue or pull request is created for the actual removal.
+    * Minor release: Compiler attributes (`OSSL_DEPRECATEDIN`) trigger
+      warnings for developers.
+    * Minor release: Support for compile-time disabling the feature 
+      (`./config no-<feature>`).
+    * Major release: Feature is compile-time disabled by default
+    * Major release: Feature moved to the Legacy Provider.
+    * Major release: Security Levels are adjusted to prohibit the feature at
+      Level 2 or higher.
+    * Record a decision that defines when removal is allowed.
     * If applicable: Adding a migration path in the ossl-migration-guide.md.
 
 ### Stage 3: OpenSSL Obsolete (Removal)
-* **Status:** Code is physically deleted.
-* **Trigger:** The next Major OpenSSL release (e.g., moving from 3.x to 4.0).
+* **Status:** Code is deleted.
+* **Trigger:** A major OpenSSL release after deprecation (e.g., moving 
+  from 3.x to 4.0).
 * **OpenSSL Action:** 
   * Removal of source code, headers, and tests.
   * Notice in CHANGES.md.
--- a/policies/obsolete-or-insecure-functionality.md
+++ b/policies/obsolete-or-insecure-functionality.md
@@ -1,0 +1,70 @@
+OpenSSL policy for obsoleting cryptographic primitives and protocols
+--------------------------------------------------------------------
+
+This formal policy outlines the OpenSSL Project’s framework for managing the
+lifecycle of cryptographic primitives, protocols, and structural features. 
+It ensures a "secure by default" posture while providing a predictable migration
+path for the global ecosystem.
+
+## 1. Policy Objective
+To maintain the integrity of the OpenSSL library by systematically deprecating
+and removing components that no longer meet modern cryptographic standards or
+industry best practices.
+
+## 2. Triggers for end-of-life
+Features enter the deprecation lifecycle when they meet one of two criteria:
+
+* **Insecure:** Primitives or features with known practical attacks, or where 
+  the "Work Factor" (computational effort to break) has decayed.
+    * Any primitive with an effective security strength below **112 bits** 
+      (OpenSSL Security Level 2) is considered insecure for modern use.
+    * Protocol features that are exploitable by design.
+* **Obsolete:** Features that remain mathematically sound but have been 
+  superseded by more secure or efficient alternatives (e.g., TLS 1.2 in the 
+  presence of TLS 1.3), or those that introduce unnecessary "state machine"
+  complexity.
+
+## 3. The end-of-life stages
+Removal of cryptographic primitives and protocols happens through several stages
+to conform to OpenSSL api guarantees.
+
+### Stage 1: Industry Obsolete (Signal)
+* **Status:** Feature is still available and enabled by default.
+* **Trigger:** An industry body (IETF, NIST, etc.) publishes a "Deprecated" or
+  "Historic" status, or a significant research paper (e.g., Logjam, CRIME) is
+  published.
+* **OpenSSL Action:** 
+  * Documentation is updated to reflect industry obsoletion status.
+  * A Github issue or pull request is created for the actual deprecation.
+
+### Stage 2: OpenSSL Deprecated (Transition)
+* **Status:** Feature is moved to the **Legacy Provider**.
+* **Trigger:** A new Major or Minor OpenSSL release.
+* **OpenSSL Action:**:
+    * Feature is disabled by default; must be explicitly loaded via configuration.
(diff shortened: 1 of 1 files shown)


levitte:
Is the legacy provider given any consideration?  I know it's been mentioned in that discussion.

The legacy provider was, in my recollection, to become a provider that retains old "obsolete" algorithms, for those that may need it (for example to decrypt something that's been at rest for a long time, or because *their* protocol still uses such an algorithm for their own reasons).

fwh-dc:
> Is the legacy provider given any consideration? I know it's been mentioned in that discussion.

It is explicitly mentioned here:
https://github.com/openssl/technical-policies/pull/118/changes#diff-561dd7e03b7ff39ada6281c8ab5d6b93053a51b3b93abd23377802c16d5b0632R41

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+--------------------------------------------------------------------
+
+This formal policy outlines the OpenSSL Project’s framework for managing the
+lifecycle of cryptographic primitives, protocols, and structural features. 
+It ensures a "secure by default" posture while providing a predictable migration
+path for the global ecosystem.
+
+## 1. Policy Objective
+To maintain the integrity of the OpenSSL library by systematically deprecating
+and removing components that no longer meet modern cryptographic standards or
```
beldmit: s/components/features/?

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+
+* **Insecure:** Primitives or features with known practical attacks, or where 
+  the "Work Factor" (computational effort to break) has decayed.
+    * Any primitive with an effective security strength below **112 bits** 
+      (OpenSSL Security Level 2) is considered insecure for modern use.
+    * Protocol features that are exploitable by design.
+* **Obsolete:** Features that remain mathematically sound but have been 
+  superseded by more secure or efficient alternatives (e.g., TLS 1.2 in the 
+  presence of TLS 1.3), or those that introduce unnecessary "state machine"
+  complexity.
```
beldmit: `introduce unnecessary "state machine" complexity.`

Too vague, I'm afraid

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+
+## 3. The end-of-life stages
+Removal of cryptographic primitives and protocols happens through several stages
+to conform to OpenSSL api guarantees.
```
beldmit: Nit: api => API

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Stage 1: Industry Obsolete (Signal)
+* **Status:** Feature is still available and enabled by default.
+* **Trigger:** An industry body (IETF, NIST, etc.) publishes a "Deprecated" or
```
beldmit: standard body?

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+  "Historic" status, or a significant research paper (e.g., Logjam, CRIME) is
```
beldmit: significant research paper demonstrating weakness

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+  published.
+* **OpenSSL Action:** 
+  * Documentation is updated to reflect industry obsoletion status.
+  * A Github issue or pull request is created for the actual deprecation.
```
beldmit: Build flags "no-feature"?

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Stage 2: OpenSSL Deprecated (Transition)
+* **Status:** Feature is moved to the **Legacy Provider**.
+* **Trigger:** A new Major or Minor OpenSSL release.
```
beldmit: Minor = 2nd digit?

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+* **OpenSSL Action:**:
+    * Feature is disabled by default; must be explicitly loaded via configuration.
```
beldmit: > must be explicitly loaded via configuration.

must be explicitly loaded via compilation flags?

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+## 4. Specific Criteria for Protocol Features
+Structural features (extensions, compression, etc.) are evaluated differently
+than mathematical primitives:
+
+1.  **Information Leakage:** Any feature that leaks plaintext length or metadata
+    (e.g., **TLS Compression**) is moved immediately to Stage 2.
+2.  **Forward Secrecy:** Features that rely on static keying material (e.g., 
+    **Static RSA**) are marked Obsolete and moved to Stage 2.
+3.  **Side-Channel Resistance:** If a feature cannot be implemented in
+    "constant-time" on modern hardware, it is prioritized for Stage 3 removal.
```
beldmit: I think it's a bit too strict.
@tomato42 WDYT?

beldmit (commented):
I think this is a great starting point but should be polished from both the text and workflow perspective

fwh-dc:
> I think this is a great starting point but should be polished from both the text and workflow perspective

I didn’t believe I would hit a home-run here. Thanks for your input!

policies/obsolete-or-insecure-functionality.md
fwh-dc: Yes exactly. I can clarify that? I thought it was standard naming convention for X.Y.Z release versioning. Major, minor, patch.

beldmit: Correct, sorry for noise

policies/obsolete-or-insecure-functionality.md
neverpanic: This text as written makes it sounds like once adopted, this document will be used to argue for the removal of TLS 1.2 from OpenSSL — which I doubt it will, but maybe we should find a different example so we don't give anybody a wrong impression.

policies/obsolete-or-insecure-functionality.md
neverpanic: I don't think this can be in a minor release, since it would break compatibility, and compatibility breaks are only OK on major releases. This should therefore be: "Trigger: A new major OpenSSL release"

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+  * A Github issue or pull request is created for the actual deprecation.
+
+### Stage 2: OpenSSL Deprecated (Transition)
+* **Status:** Feature is moved to the **Legacy Provider**.
+* **Trigger:** A new Major or Minor OpenSSL release.
+* **OpenSSL Action:**:
+    * Feature is disabled by default; must be explicitly loaded via configuration.
+    * Compiler attributes (`OSSL_DEPRECATEDIN`) trigger warnings for developers.
+    * Security Levels are adjusted to prohibit the feature at Level 2 or higher.
+    * An issue or pull request is created for the actual removal.
```
neverpanic: I'm not sure that's worthwhile — it could be a year or more before the code actually gets removed, at which point the pull request will be stale and will essentially have to be re-done.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+    * If applicable: Adding a migration path in the ossl-migration-guide.md.
+
+### Stage 3: OpenSSL Obsolete (Removal)
+* **Status:** Code is physically deleted.
```
neverpanic:
```suggestion
* **Status:** Code is deleted.
```
I'm not quite sure what the difference between physically deleting code and non-physically (virtually?) deleting code is, so maybe let's just drop the qualifier.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,70 @@
...
+* **Trigger:** The next Major OpenSSL release (e.g., moving from 3.x to 4.0).
```
neverpanic: I think this should read "the next major OpenSSL release after deprecation", and there should be some wiggle room here in case we want to keep things for a while longer due to widespread use.

Let's say OpenSSL deprecates TLS 1.2 and disables it by default, first maybe using configuration, then using a build flag — we may want to keep the build flag for 5 years or more before actually removing the code just because how widely used 1.2 still is.

policies/obsolete-or-insecure-functionality.md
fwh-dc: Yeah I get that. Actually the intent here was that TLS 1.2 is considered obsolete from an industry point of view i.e. the TLS 1.3 RFC. So actually by definition we should begin considering what to do about the support now.

But I will just remove this part as it is superfluous.

policies/obsolete-or-insecure-functionality.md
fwh-dc: You are right. Updated.

policies/obsolete-or-insecure-functionality.md
fwh-dc: Yes, I'm thinking of how and where to document these decisions. I was thinking of a table of all supported but deprecated/insecure features and then an estimate timeline for deprecation and removal.

But I agree this statement is too strict.

policies/obsolete-or-insecure-functionality.md
fwh-dc: Agreed I've rephrased this to avoid the "how it is done".

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
+OpenSSL policy for obsoleting cryptographic primitives and protocols
```
t8m:
```suggestion
OpenSSL Library policy for obsoleting cryptographic primitives and protocols
```

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+--------------------------------------------------------------------
+
+This formal policy outlines the OpenSSL Project’s framework for managing the
```
t8m: Project -> Library as different projects under the OpenSSL umbrella can have a different policy.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+lifecycle of cryptographic primitives, protocols, and structural features. 
+It ensures a "secure by default" posture while providing a predictable migration
+path for the global ecosystem.
+
+## 1. Policy Objective
+To maintain the integrity of the OpenSSL library by systematically deprecating
```
t8m:
```suggestion
To maintain the integrity of the OpenSSL Library by systematically deprecating
```

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+## 2. Triggers for end-of-life
+Features enter the deprecation lifecycle when they meet one of two criteria:
+
+* **Insecure:** Primitives or features with known practical attacks, or where 
+  the "Work Factor" (computational effort to break) has decayed.
+    * Any primitive with an effective security strength below **112 bits** 
+      (OpenSSL Security Level 2) is considered insecure for modern use.
+    * Protocol features that are exploitable by design.
+* **Obsolete:** Features that remain mathematically sound but have been 
+  superseded by more secure or efficient alternatives.
```
t8m: This is too vague. What does it mean superseded. TLS-1.2 is superseded by TLS-1.3 but we are not going to deprecate it until its use diminishes (or it becomes insecure -> handled by the first option).
```suggestion
* **Obsolete:** Features that remain cryptographically sound but have no 
practical use anymore.
```

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+to conform to OpenSSL API guarantees.
+
+### Stage 1: Industry Obsolete (Signal)
+* **Status:** Feature is still available and enabled by default.
+* **Trigger:** An industry standards body (IETF, NIST, etc.) publishes a
+  "Deprecated" or "Historic" status, or a significant research paper 
+  demonstrating a weakness (e.g., Logjam, CRIME) is published.
+* **OpenSSL Action:** 
+  * Documentation is updated to reflect industry obsoletion status.
+  * Record a decision that defines when deprecation is allowed.
```
t8m: I think `planned` would be better word than `allowed` here.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Stage 2: OpenSSL Deprecated (Transition)
+* **Status:** Feature is moved to the **Legacy Provider**.
```
t8m: This does not apply to things other than crypto algorithms. Protocols are not in providers.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+* **Trigger:** A new major or minor OpenSSL release.
+* **OpenSSL Action:**:
+    * Minor release: Compiler attributes (`OSSL_DEPRECATEDIN`) trigger
```
t8m, line 52: This does not apply to provided functionality as there are no APIs that can be deprecated.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+
+## 3. The end-of-life stages
+Removal of cryptographic primitives and protocols happens through several stages
+to conform to OpenSSL API guarantees.
+
+### Stage 1: Industry Obsolete (Signal)
+* **Status:** Feature is still available and enabled by default.
+* **Trigger:** An industry standards body (IETF, NIST, etc.) publishes a
+  "Deprecated" or "Historic" status, or a significant research paper 
+  demonstrating a weakness (e.g., Logjam, CRIME) is published.
```
t8m, line 33: There are very much different levels of weakness and historic/legacy statuses for things. That something has a weakness does not necessarily mean that it should be immediately deprecated or even moved to a different provider at next minor/major release. We have to compare the severity of the weakness with the existing deployment of the technology. It really has to be judged based on multiple things.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+* **OpenSSL Action:**:
+    * Minor release: Compiler attributes (`OSSL_DEPRECATEDIN`) trigger
+      warnings for developers.
+    * Minor release: Support for compile-time disabling the feature 
+      (`./config no-<feature>`).
+    * Major release: Feature is compile-time disabled by default
+    * Major release: Feature moved to the Legacy Provider.
+    * Major release: Security Levels are adjusted to prohibit the feature at
+      Level 2 or higher.
+    * Record a decision that defines when removal is allowed.
```
t8m: Again `allowed` -> `planned`.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+* **Trigger:** A new major or minor OpenSSL release.
+* **OpenSSL Action:**:
+    * Minor release: Compiler attributes (`OSSL_DEPRECATEDIN`) trigger
+      warnings for developers.
+    * Minor release: Support for compile-time disabling the feature 
+      (`./config no-<feature>`).
+    * Major release: Feature is compile-time disabled by default
+    * Major release: Feature moved to the Legacy Provider.
+    * Major release: Security Levels are adjusted to prohibit the feature at
+      Level 2 or higher.
```
t8m: This can be done at minor release if there is a good reason.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+    * Record a decision that defines when removal is allowed.
+    * If applicable: Adding a migration path in the ossl-migration-guide.md.
+
+### Stage 3: OpenSSL Obsolete (Removal)
+* **Status:** Code is deleted.
+* **Trigger:** A major OpenSSL release after deprecation (e.g., moving
```
t8m, line 57: IMO not necessarily. I can see that something gets deprecated and stays there for quite some time. Especially if we will have major releases more often that we currently do. We will always have to consider industry usage (for legacy purposes) before removing something completely.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+* **OpenSSL Action:** 
+  * Removal of source code, headers, and tests.
+  * Notice in CHANGES.md.
+
+## 4. Specific Criteria for Protocol Features
+Structural features (extensions, compression, etc.) are evaluated differently
+than mathematical primitives:
+
+1.  **Information Leakage:** Any feature that leaks plaintext length or metadata
+    (e.g., **TLS Compression**) is moved immediately to Stage 2.
```
t8m: Huh? No. We are not going to deprecate things in patch releases.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+2.  **Forward Secrecy:** Features that rely on static keying material (e.g., 
+    **Static RSA**) are marked Obsolete and moved to Stage 2.
```
t8m: Same as above.

policies/obsolete-or-insecure-functionality.md
t8m: This is basically asking for removal of RSA encryption. IMO we are not removing it in 5.0.

I would remove the whole section 4 as it does not make much sense to me at this point.

policies/obsolete-or-insecure-functionality.md
```diff
@@ -0,0 +1,74 @@
...
+3.  **Side-Channel Resistance:** If a feature cannot be implemented in
+    "constant-time" on modern hardware, it is prioritized for Stage 3 removal.
+
+> **Emergency Override:** The OpenSSL Foundation Board of Directors reserves the
+> right to bypass Stages 1 and 2 for "Zero-Day" catastrophic breaks, disabling
+> the feature in a minor security patch.
```
t8m: It would really take some truly catastrophic break to bypass Stage 1. I cannot imagine bypassing Stage 2.
It could be that the break would enforce doing a major version earlier than scheduled. So there would be a major version instead of a minor one done earlier where the broken algorithm would be moved to legacy or some protocol feature was deprecated and disabled by default.

t8m (changes requested):

policies/obsolete-or-insecure-functionality.md
levitte: > This is basically asking for removal of RSA encryption.

Would it?  Don't we at least have the *ambition* of constant-time on per key size basis?

(Here, I echo something DJB said at ...  I think it was ICMC in Vancouver ..., that you can view each RSA key size as a separate algorithm, and that for each of size, constant-time should be possible to achieve)

t8m: No, this is not because the keys have variable size. The private key RSA decryption with PKCS1 1.5 padding is basically impossible to do without providing an oracle to the attacker unless workarounds like those in TLS or the implicit rejection is applied. So yeah, unless you consider these workarounds as _fixes_, we would have to remove the support.

Similarly for example for AES unless implemented in hardware or in assembler as the truly constant time implementation in pure C is so slow that it is unusable for many applications. Should this be removed forcibly?

policies/obsolete-or-insecure-functionality.md
fwh-dc: This is the second time that I need to explain this bit so definitely this needs to be improved!

The intention here is to say for example:

> When a standardisation body standardises TLS 1.3 then it obsoletes TLS 1.2. Thereby TLS 1.2 enter the first stage of end-of-life which is stage 1: Industry obsoletion. This should trigger a response from the OpenSSL library i.e. promotion of TLS 1.3 over TLS 1.2 for new applications (through documentation and example code), maybe also providing migration paths for existing applications.
> The OpenSSL board of directors (I'm guessing this is the replacement for OMC) should then record a decision about deprecation: either at next possible opportunity or wait for industry migration from TLS 1.2 to a newer standard.

Please let me know if you believe this is a wrong path.

policies/obsolete-or-insecure-functionality.md
fwh-dc: I agree with you. So obviously this needs to be updated to reflect the point I am trying to make better. I'm thinking...

policies/obsolete-or-insecure-functionality.md
fwh-dc: patch release? you mean minor releases? Point here is that the library is not waiting for an industry standards body to deprecate the feature, but can go ahead with some of the steps during deprecation. For example implementing support for disabling the feature at compile time.

policies/obsolete-or-insecure-functionality.md
fwh-dc: Probably this section is a bit farfetched. I can leave it out.

policies/obsolete-or-insecure-functionality.md
fwh-dc: I'm not proposing immediate deprecation. I'm proposing taking a decision of what should happen.

policies/obsolete-or-insecure-functionality.md
t8m: I do not think there is any action to be done in regards to the older version on our side when a new version is standardized. That is way too early to decide when TLS-1.2 will enter the deprecation and later removal stages. That something new appears does not necessarily mean that the old thing that it replaces is already obsolete. It might (and usually will) have its use cases which are still sound and perfectly OK to be used.

Yes, TLS-1.2 is now entering the stage when we could start thinking about its obsolescence but it is not because TLS-1.3 appeared (that is a necessary but definitely not sufficient condition), but because the Quantum Computing threat is starting to get real and TLS-1.2 does not really have any answer to that threat.

policies/obsolete-or-insecure-functionality.md
baentsch: Coming here from the Community discussion (thanks for the pointer @fwh-dc and this write-up!) and taking the statements there into account (that apparently there's no universal appetite to even remove proven-insecure stuff as there _may_ be someone relying on it -- the "poured concrete" argument), could it be that this stage can never be reached? And if so, that the previous stage should be formulated accordingly? Something along the lines 
- Make it _really_ difficult to enable the feature.
- Display warning in BOLD (at least in documentation and at compile time) that such deprecated feature is no longer "wise to use". 
- Maybe even output such warning at every execution if not consciously "configured away" by someone (e.g., distro maintainer) knowing (and accepting) the risks?
- Added idea: Anyone enabling such "strongly discouraged" feature might be encouraged to send an email to an automated "feature-use tracking server" -- whose data then may be used to ultimately decide about actual feature removal after all: Just maybe, the floor can eventually be lifted after all based on such definite "(still|not)in-use indicators"?

t8m (approved):

PR #120, opened 2026-06-16
Update assembler policy to mention compiler intrinsic code

mbroz:
OpenSSL already uses compiler intrinsic at some places, let's add mention of it to the Assembler policy.

### DIFF
--- a/policies/assembler.md
+++ b/policies/assembler.md
@@ -14,6 +14,16 @@ acceptable where such updates would be allowed under the
 Where assembler optimisations are acceptable they should be implemented using
 [perlasm].
 
+Compiler intrinsic code can be used where it makes sense.
+If compiler support for intrinsic code is not present, it must be automatically
+disabled during the build (for compilers that are still supported).
+
+Selection of an implementation for an architectural feature must always be
+based on runtime detection of the CPU's capabilities.
+It is not acceptable to fail during runtime due to missing CPU capabilities
+unless the build configuration explicitly targets the concrete CPU family
+that provides the capability.
+
 [perlasm]: /policies/general/glossary/#perlasm
 [stable release]: /policies/general/glossary/#stable-release
 [stable release update policy]: /policies/technical/stable-release-updates/


t8m (approved):

t-j-h (approved):

policies/assembler.md
```diff
@@ -14,6 +14,9 @@ acceptable where such updates would be allowed under the
 Where assembler optimisations are acceptable they should be implemented using
 [perlasm].
 
+In exceptional cases, compiler intrinsic code can be used.
```
tom-cosgrove-arm: But what would justify use? What do we mean by "exceptional"?

Personally, I am in favour of intrinsics - let the compiler do the loop unrolling and space/performance trade-offs without us having to provide multiple almost-identical implementations.

policies/assembler.md
```diff
@@ -14,6 +14,9 @@ acceptable where such updates would be allowed under the
...
+All supported toolchains for the particular platform must support it.
```
tom-cosgrove-arm: Is that even possible? What we've typically done is put #ifdef guards in place so that the build will always work, even if it's not the most performant on (for example) an older compiler that doesn't support a newer intrinsic.

paulidale: Could we have some standard guards just for this purpose?

```c
#ifdef OSSL_HAS_VECTOR256_INTRINSICS
...
#else
...
#endif
```

policies/assembler.md
```diff
@@ -14,6 +14,9 @@ acceptable where such updates would be allowed under the
...
+
```
t8m:
```suggestion
Compiler intrinsic code can be used where it makes sense.

Selection of the implementation to be used on a particular architecture must
always be done based on a run-time detection of the CPU capabilities.
It is not acceptable to fail during run-time due to missing CPU capabilities
unless the build configuration target explicitly targets the concrete CPU
family that provides the capability and the execution happens on a different
CPU family.

```

t8m (commented):
What about this change?

policies/assembler.md
mbroz: Applied, just I used slightly shorter wording here.

policies/assembler.md
mbroz: I meant is as perlasm is preferred, but we replaced it with "Compiler intrinsic code can be used where it makes sense."

policies/assembler.md
mbroz: This should go into coding style. After discussion with @t8m we added paragraph about CPU capabilities.

policies/assembler.md
```diff
@@ -14,6 +14,14 @@ acceptable where such updates would be allowed under the
 Where assembler optimisations are acceptable they should be implemented using
 [perlasm].
 
+Compiler intrinsic code can be used where it makes sense.
```
tom-cosgrove-arm, line 17: What about building if the compiler doesn't support the intrinsics. I think we should call out that builds mustn't fail, but it should be acceptable that specific compiler(s)/version(s) might be needed to get the benefit

policies/assembler.md
```diff
@@ -14,6 +14,14 @@ acceptable where such updates would be allowed under the
...
+
+Selection of an implementation for a particular architecture must always be
```
tom-cosgrove-arm: > a particular architecture

Err, this reads as if could should determine if it's running on x86_64 or AArch64 at runtime :)

Maybe we should say "architectural feature"

policies/assembler.md
mbroz: I think that no-asm or something similar has to be used for build then.

policies/assembler.md
mbroz: Ah right, sounds strange. Will change it.

mbroz: fixed

policies/assembler.md
t8m, line 17:
```suggestion
Compiler intrinsic code can be used where it makes sense. If the compiler support
for them is not universally present on the target platform, they must be automatically
disabled during the build so the build does not fail on compilers which are still
supported but do not support the particular intrisics.
```

t8m: What about this?

mbroz: The question is, what is a compiler still supported...

mbroz: I added simplified version, please check.

tom-cosgrove-arm: Hmm, I disagree there. Intrinsics are C - they can be protected by #includes with suitable version guards. The point is, we should be explicit that PRs adding intrinsics mustn't break builds even on unusual compilers

tom-cosgrove-arm (approved):
LGTM, thanks

policies/assembler.md
```diff
@@ -14,6 +14,16 @@ acceptable where such updates would be allowed under the
 Where assembler optimisations are acceptable they should be implemented using
 [perlasm].
 
+Compiler intrinsic code can be used where it makes sense.
+If compiler support for intrinsic code is not present, it must be automatically
+disabled during the build (for compilers that are still supported).
```
beldmit, line 19: Must be guarded by corresponding `ifdefs`?

t8m: That is IMO unnecessarily specific.

t8m (approved):
Approving just formally as this is basically my proposal. @t-j-h @arapov @mattcaswell ping for approvals.

arapov (approved):

t-j-h (approved):

t8m:
Merged. Thank you.

